PR #71, opened 2026-05-22
fix: update Linux runtime library path handling

yizhinailong:
- Update comment to clarify that runtime_lib_dirs() provides Linux toolchain runtime library search paths instead of generic Linux-specific paths
- Remove hardcoded system library paths (/lib/x86_64-linux-gnu, /usr/lib/x86_64-linux-gnu, /usr/lib64) from runtime_lib_dirs() function to rely solely on toolchain-specific paths


Sunrisepeak (approved):
lgtm

PR #115, opened 2026-06-04
[temp][DO NOT MERGE] macOS 14 support experiment CI

Sunrisepeak:
## 目的（临时 PR，勿合并）

调研并验证 mcpp + xlings 对 **macOS 14** 的支持路径。

## 现状（调研结论）

- 已发布的 macosx-arm64 产物（mcpp 0.0.49 / xlings 0.4.49）`LC_BUILD_VERSION minos=15.0 sdk=15.0` —— **macOS 14 的 dyld 直接拒载**。根因：macos-15 runner 构建时未设 deployment target，取了 SDK 默认。
- 两个二进制都动态链接系统 `/usr/lib/libc++.1.dylib` —— 降 target 需要通过 libc++ availability 门控（编译期）且无运行期缺符号（本实验验证点）。
- mcpp 构建链（flags.cppm）无任何 deployment target 处理 → clang driver 尊重 `MACOSX_DEPLOYMENT_TARGET` 环境变量，机制上环境变量即可生效。

## 实验设计

| job | runner | 内容 |
|-----|--------|------|
| build-minos14 | macos-15 | bootstrap mcpp → `MACOSX_DEPLOYMENT_TARGET=14.0` 构建 mcpp@HEAD 与 xlings@main → 断言 minos=14.0 → 15 上冒烟 |
| smoke-macos14 | macos-14 | 对照：0.0.49 产物应被 dyld 拒载；minos14 产物 `--version` 启动；组装 xlings home 后 `mcpp new/build/run` 全链路（在 14 上安装 llvm/ninja 工具链包）|

验证通过后将把 deployment target 固化进两 repo 的 release/CI 并发布新版本。

PR #116, opened 2026-06-04
feat(0.0.50): macOS min-version support — static LLVM libc++, lld link, explicit deployment target

Sunrisepeak:
## Goal

macosx-arm64 release binaries should run on **macOS 11.0+** (first Apple Silicon release) instead of only macOS 15. Verified end-to-end on CI (#115, runs A→B5).

## Evidence chain (temp PR #115)

| run | config | result |
|-----|--------|--------|
| A | minos 14 + dynamic system libc++ | compiles, **dies at launch on macOS 14.8**: `dyld: Symbol not found: __ZNSt3__119__is_posix_terminalEP7__sFILE` (std::print support symbol; macOS 14's libc++ predates LLVM 18) — lowering minos alone is not viable |
| B2/B3 | `[build] ldflags` injection of static archives | minos 11 OK but still dynamically linked: the macOS link path hardcodes ` -lc++` **before** user ldflags |
| B initial | mixed deployment targets in one sandbox | `std.pcm` config-mismatch — BMI fingerprint didn't cover the deployment target |
| B4 | native staticStdlib (this PR), two-stage self-host | static + minos 11 ✅, but macos-14 e2e hit Xcode 15.4's system `ld` aborting at launch (its libc++ resolution diverted to the LLVM payload dylib, missing `__ZdaPv`) |
| **B5** | + `-fuse-ld=lld` | **all green**: macos-15 two-stage build (minos 11.0, no libc++ dylib dep, clean exit) + macos-14 control/refusal, binary startup, full `mcpp new/build/run` incl. llvm/ninja install |

## Changes

- **flags.cppm** — implement `staticStdlib` (the manifest default, previously silently ignored on the clang route) for the macOS link path: `-nostdlib++ <llvm>/lib/libc++.a <llvm>/lib/libc++abi.a` instead of `-lc++` (falls back when archives absent); link via `-fuse-ld=lld` (same as the Linux clang path — removes the host-Xcode `ld` dependency); mirror `MACOSX_DEPLOYMENT_TARGET` onto compile+link command lines.
- **cli.cppm** — fold `MACOSX_DEPLOYMENT_TARGET` into the BMI fingerprint (deployment target changes the effective compile triple; stale `std.pcm` reuse otherwise hard-fails).
- **main.cpp** — `__APPLE__` `_Exit` guard after stream flush (static libc++ static-dtor abort prevention; same guard xlings uses).
- **release.yml (macos job)** — `MACOSX_DEPLOYMENT_TARGET=11.0` + two-stage self-host (stage 1 bootstrap-built dynamic → stage 2 rebuilds itself static) + minos/no-dylib assertions.
- version 0.0.50.

## Design doc

xlings `.agents/docs/2026-06-05-macos-min-version-support.md` (kept updated through the experiments).

## Compatibility

- Linux/Windows paths untouched (`needs_explicit_libcxx` branch is macOS-only; gcc `-static-libstdc++` behavior unchanged).
- User projects on macOS now also link static libc++ by default (staticStdlib default) and via lld — verified by the macos-14 `mcpp new/build/run` e2e. Host-SDK-default minos for user binaries unchanged.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -326,8 +326,13 @@ jobs:
           MCPP=$(cd "$(dirname "$MCPP")" && pwd)/$(basename "$MCPP")
           "$MCPP" self config --mirror GLOBAL
           "$MCPP" test || {
-            echo "=== TEMP forensics: retry verbose (remove before merge) ==="
-            "$MCPP" test -v 2>&1 | grep -B3 -A12 -iE "error|undefined|ld64" | head -120
+            echo "=== TEMP forensics: retry verbose unfiltered (remove before merge) ==="
+            vm_stat | head -5
+            "$MCPP" test -v > /tmp/mtv.log 2>&1
+            echo "--- verbose tail ---"
+            tail -200 /tmp/mtv.log
+            echo "--- ninja FAILED lines ---"
+            grep -B2 -A20 "FAILED" /tmp/mtv.log | head -80
             exit 1
           }
 
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -325,7 +325,11 @@ jobs:
           MCPP=$(find target -path "*/bin/mcpp" | head -1)
           MCPP=$(cd "$(dirname "$MCPP")" && pwd)/$(basename "$MCPP")
           "$MCPP" self config --mirror GLOBAL
-          "$MCPP" test
+          "$MCPP" test || {
+            echo "=== TEMP forensics: retry verbose (remove before merge) ==="
+            "$MCPP" test -v 2>&1 | grep -B3 -A12 -iE "error|undefined|ld64" | head -120
+            exit 1
+          }
 
       - name: "TEMP forensics: static libc++ exit-abort matrix (remove before merge)"
         if: always()
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -325,39 +325,7 @@ jobs:
           MCPP=$(find target -path "*/bin/mcpp" | head -1)
           MCPP=$(cd "$(dirname "$MCPP")" && pwd)/$(basename "$MCPP")
           "$MCPP" self config --mirror GLOBAL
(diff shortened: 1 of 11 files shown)


Sunrisepeak:
Added on top: `[build] macos_deployment_target` — first-class manifest field for the minimum supported macOS version of produced binaries.

Design follows ecosystem conventions: `MACOSX_DEPLOYMENT_TARGET` env (the knob cargo/rustc/cc honor) remains the explicit per-invocation **override**; the manifest field is the **project default** (SwiftPM `platforms:` style); toolchain/SDK default otherwise. The resolved value flows into the explicit `-mmacosx-version-min` compile/link flags and the BMI fingerprint introduced earlier in this PR. Unit tests + `docs/05-mcpp-toml.md` included.

```toml
[build]
macos_deployment_target = "11.0"
```

Sunrisepeak:
## Deferred work registry (tracked, not in this PR)

Marked as `TODO(...)` in code and registered in the design doc (xlings `.agents/docs/2026-06-05-macos-min-version-support.md` §5):

| # | item | code marker | unblocks |
|---|------|-------------|----------|
| 1 | Mixed C/C++ static binaries SIGSEGV at runtime (e2e `36_llvm_toolchain`: `answer.c` + `std::cout` main, exit 139; pure-C++ fine; root cause not yet isolated) | `flags.cppm TODO(macos-static-default)` | flipping static libc++ to the unconditional default |
| 2 | std-module staging vs fingerprint boundary — a built-in default floor injected into canonical flags alone leaves the test build's std.pcm unstaged (`import std` fails wholesale). Fix: one resolver shared by flags/fingerprint/stdmod | `cli.cppm TODO(macos-default-floor)` | re-landing the rustc-style built-in default floor |
| 3 | Custom libc++ build (floor 11.0–13.0) via xlings-res; data-only swap of the archive source | `flags.cppm TODO(macos-floor-11)` | floor below 14 |

End state when #1+#2 land: static libc++ + built-in 14.0 floor become the unconditional macOS defaults (cargo-style portable-by-default); the declared-floor gate stays only as the opt-out boundary.

PR #119, opened 2026-06-05
refactor(macos): single deployment-target resolver (flags + fingerprint + stdmod)

Sunrisepeak:
First 0.0.51 change (deferred TODO #2 from the PR #116 registry).

`platform::macos::deployment_target(manifestValue)` becomes the single resolution point (env `MACOSX_DEPLOYMENT_TARGET` override > `[build] macos_deployment_target` manifest default > empty), consumed by:

- `flags.cppm` — compile + link `-mmacosx-version-min`
- the BMI fingerprint rule (`cli.cppm`)
- **the std-module prebuild** (`stdmod.cppm ensure_built`, new parameter) — previously stdmod assembled its compile flags with no deployment-target concept, so a manifest-resolved target split the `std.pcm` triple from the importing TUs (`import std` failed wholesale on macos CI when a built-in default was injected into the fingerprint alone). Env-only setups merely got lucky via the clang driver reading the env in both places.

Unblocks re-landing the rustc-style built-in default floor (`TODO(macos-default-floor)`). No behavior change for env-only or unconfigured builds. 18/18 unit tests pass.

### DIFF
--- a/src/build/flags.cppm
+++ b/src/build/flags.cppm
@@ -103,12 +103,8 @@ CompileFlags compute_flags(const BuildPlan& plan) {
     // override, the convention cargo/rustc/cc honor) > the manifest's
     // [build] macos_deployment_target (project default, SwiftPM-style) >
     // empty (toolchain/SDK default).
-    std::string macosDeploymentTarget;
-    if (const char* dt = std::getenv("MACOSX_DEPLOYMENT_TARGET"); dt && *dt) {
-        macosDeploymentTarget = dt;
-    } else {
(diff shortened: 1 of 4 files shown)


Sunrisepeak:
CI: macos + windows green. The linux job fails only at e2e `29_toolchain_partial_versions.sh` / `31_transitive_deps.sh` — the identical, deterministic registry-side failure already present on main (run 26988545086) and tracked in #120 (gcc@16.1.0 first-run install: glibc 2.39 xpkg lacks `linux/limits.h`). Unrelated to this change; merging with admin bypass.

Issue #243, opened 2026-07-18
[enhancement] manifest [features] 支持条件依赖与依赖 feature 转发(dep/feat;挡住 opencv 模块包的 opencv.dnn 接口)


Sunrisepeak:
## 诉求

mcpp.toml(manifest)侧的 `[features]` 目前支持 sources/implies/defines,
但不支持:
1. **feature 条件依赖**(描述符 lua 侧 0.0.97 已支持 `features.x.deps`,
   manifest 侧没有对应物);
2. **向依赖转发 feature**(cargo 的 `dep/feat` 语法对应物):
   `features.dnn = { deps-features = { "compat.opencv" = ["dnn"] } }` 之类。

## 场景(opencv-m 的 opencv.dnn 模块接口,mcpp-index PR #79 后续)

compat.opencv 已带 `dnn` feature(+309 TU)。opencv 模块包想提供
`import opencv.dnn;`,理想形态:

```toml
[features]
dnn = { sources = ["src/dnn.cppm"], forward = { "compat.opencv" = ["dnn"] } }
```

消费端 `opencv = { version = "0.0.3", features = ["dnn"] }` → 模块包启用
dnn.cppm 源 + 依赖的 compat.opencv 以 features=["dnn"] 解析。没有 (2),
模块包只能永远全量拉 dnn(所有消费者多 300 TU)或不提供 dnn 模块接口。

## 环境

mcpp 0.0.97

Sunrisepeak:
**Status (0.0.98):** design finalized — see `.agents/docs/2026-07-19-issue-243-feature-forwarding-design.md`. Scoping clarified: feature-conditional deps ALREADY exist (`[feature-deps]` / xpkg `features.x.deps`); the real gap is `dep/feat` **forwarding**. Implementation is a follow-up PR (0.0.99+) — the #242 work already converged the per-package feature request onto `DependencyEdge`/`aggregatedRequest`, which the forwarding builds on. Keeping open for the impl.

Sunrisepeak:
Resolved in **mcpp 0.0.99** (released + indexed).

Feature `dep/feat` forwarding is implemented (Cargo parity). A `[features]` implied token containing `/` — or the dedicated table-form `forward = ["dep/feat"]` key — forwards a feature to a dependency:

```toml
[features]
dnn = { sources = ["src/dnn.cppm"], implies = ["compat.opencv/dnn"] }
[dependencies]
compat.opencv = { version = "0.0.3", default-features = false }
[...]

Issue #230, opened 2026-07-17
[bug][windows] 0.0.95 下 mcpp test --workspace 以 exit 127 中断(0.0.94 同负载全绿)


Sunrisepeak:
## 现象

mcpp-index 的 `workspace (windows)` CI job(windows-latest,vendored xlings,`mcpp test --workspace`,17 个成员)在 **0.0.95** 下于成员依赖下载阶段中断,step 以 **exit 127**(command not found)失败;同一成员集在 **0.0.94**(main 分支每日 CI)持续全绿。macOS/linux 同 run 无此现象(linux 的失败另有原因,见 mcpp-index#74)。

## 证据

- 失败 run(mcpp-index PR #74,pin bump 0.0.94→0.0.95):
  https://github.com/mcpplibs/mcpp-index/actions/runs/29614755884/job/87997292043
  - 21:27:45 起正常下载多个 compat 包(zstd 等)
  - 21:27:53-55 下载 compat.libarchive@3.8.7 进行中,21:27:55 step 直接
    `##[error]Process completed with exit code 127`
- 对照:mcpp-index main(0.0.94 pin)validate 每日 success(2026-07-17 等)。

## 初步怀疑

0.0.95 引入的三个特性中,windows 路径上新增子进程/工具调用的是 #222(依赖包 build.mcpp 执行)与 #220(nasm 解析);127 指示某个被 spawn 的命令在 windows runner 上不存在。workspace 成员含 `tests/examples/build-mcpp`(带 build.mcpp),0.0.95 起依赖/成员的 build.mcpp 在 windows 的执行细节(shell/驱动路径)值得优先排查。

## 影响与临时对策

阻塞 mcpp-index 的 MCPP_VERSION 升级(compat.ffmpeg 描述符需要 0.0.95 的 per-glob flags 语法)。临时对策:windows 矩阵腿单独钉回 0.0.94(该平台不消费新语法,cfg(linux) 门控成员为 no-op),linux/macos 用 0.0.95;待本 issue 修复后统一。

Sunrisepeak:
根因已定位并修复:见 #231。

[...]

PR #244, opened 2026-07-18
fix(build): object-path disambiguation follow-ups #239 #240 (0.0.98)

Sunrisepeak:
Fixes the two #233 object-path disambiguation follow-up gaps that block mcpplibs #79 (opencv intake).

- **#240**: after disambiguation the entry-main link input still referenced the stale flat `obj/main.o`; converged object-path assignment onto a single `object_for` so link input == compile edge (e2e 123).
- **#239**: out-of-root / absolute relPath (dep build.mcpp OUT_DIR sources) escaped `obj/`, and the `@` in `<name>@<ver>` broke the #235 depfile redirect; `safe_object_prefix` folds every component downward + shell-safe (e2e 124).

Common single-binary projects are byte-identical. Design + ledger: `.agents/docs/2026-07-19-*`.
Unblocks #79 (the merge gate). Ship as 0.0.98.


Sunrisepeak:
Superseded by #245 — the whole #230–#243 batch is unified into a single 0.0.98 release there (#239/#240 included).

Issue #265, opened 2026-07-21
mcpp new --template imgui fails on all platforms: package 'imgui@1.92.8' has no mcpp.toml


Sunrisepeak:
## Summary

`mcpp new <name> --template imgui` fails on **all three platforms** in the
post-release / scheduled `ci-fresh-install` workflow:

```
 Downloading compat.imgui v1.92.8
error: package 'imgui@1.92.8' has no mcpp.toml
```

The step is `Template: mcpp new --template imgui (fetch path)`, and it fails
identically on `Linux fresh install`, `macOS fresh install` and
`Windows fresh install`.

## Not a 0.0.102 regression

Found while verifying the 0.0.102 release, but it predates it. The error text
is byte-identical in the scheduled run that executed **before** the 0.0.102
batch landed:

| run | date | mcpp | result |
|---|---|---|---|
| [29736516661](https://github.com/mcpp-community/mcpp/actions/runs/29736516661) | 2026-07-20 | 0.0.101 release | all jobs green |
| [29814932554](https://github.com/mcpp-community/mcpp/actions/runs/29814932554) | 2026-07-21 08:36 | scheduled | `error: package 'imgui@1.92.8' has no mcpp.toml` |
| [29873685262](https://github.com/mcpp-community/mcpp/actions/runs/29873685262) | 2026-07-21 22:30 | 0.0.102 release | same error, same package, same version |

So the regression window is between 2026-07-20 and 2026-07-21 — after
0.0.101's post-release verification passed and before the July 21 schedule.
Nothing in mcpp released in that window, which points at the package/index
side (`compat.imgui` 1.92.8 or the template's fetch path) rather than at the
engine.

## Where to look

- `compat.imgui` at 1.92.8: does the payload actually carry an `mcpp.toml`,
  or is it expected to be synthesized from the descriptor's `mcpp = { ... }`
  segment? The message comes from the template/scaffold fetch path, which
  looks for a manifest inside the fetched package.
- Whether the template flow and the ordinary dependency flow disagree about
  which of those two forms is acceptable — an ordinary `[dependencies]` use
  of `compat.imgui` is not reported failing, which would make this specific
  to the scaffold path.

## Secondary observation in the same runs

The same jobs print

```
error: index requires mcpp >= 0.0.101 but this is mcpp 0.0.100 [E0006]
```

That one was the workspace bootstrap pin in `.xlings.json` lagging the index
[...]

PR #266, opened 2026-07-22
fix: skip Form B packages in template name resolution

xv1rcn:
## Summary

- `fetch_template_package()` now skips Form B packages whose manifest
  is synthesized from the descriptor (`mcpp = { ... }`)
- Uses existing `mcpp::manifest::extract_mcpp_field()` API, consistent
  with `prepare.cppm` Form A/B dispatch
- 4 lines added in `src/scaffold/create.cppm`

Refs #265 (hardening — does not close it; see review comment below for the
verified root cause, which lands in a separate PR)

## Test plan

- [x] `bash tests/e2e/69_package_templates.sh` — new Form B filter case
- [x] `bash tests/e2e/02_new_build_run.sh` — no regression
- [x] `bash tests/e2e/12_add_command.sh` — no regression
- [x] `bash tests/e2e/27_namespace_dependencies.sh` — no regression
- [x] CI: Linux / macOS / Windows

### DIFF
--- a/src/scaffold/create.cppm
+++ b/src/scaffold/create.cppm
@@ -44,6 +44,12 @@ fetch_template_package(const mcpp::scaffold::TemplateSpec& spec) {
     std::optional<std::string> lua;
     for (std::string cand : {std::string{}, std::string{"compat"}}) {
         if (auto l = fetcher.read_xpkg_lua(cand, spec.pkg)) {
+            // Form B packages (mcpp = { ... }) are build recipes without
+            // a source tree — no physical mcpp.toml or templates/.
+            auto field = mcpp::manifest::extract_mcpp_field(*l);
+            if (field.kind == mcpp::manifest::McppField::TableBody) {
+                continue;
+            }
             lua = std::move(*l);
             break;
         }
--- a/tests/e2e/69_package_templates.sh
+++ b/tests/e2e/69_package_templates.sh
@@ -133,4 +133,31 @@ grep -qi "not found in the index" out5.log || { cat out5.log; echo "missing inde
 "$MCPP" new app6 --template gui > out6.log 2>&1 || { cat out6.log; echo "gui builtin broke"; exit 1; }
 grep -qi "deprecated" out6.log || { cat out6.log; echo "missing gui deprecation"; exit 1; }
 
+# 7. Form B (mcpp = { ... }) compat package → skipped, "not found in the index".
+mkdir -p "$MCPP_HOME/registry/data/mcpplibs/pkgs/c"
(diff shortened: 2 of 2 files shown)


Sunrisepeak:
Thanks — reviewed and reproduced locally. The Form B filter is worth having, but it is **not** what makes #265 fail, so I changed `Closes #265` → `Refs #265`.

### What actually flips the resolution

Not "`imgui.lua` unavailable for some reason" — the index floor gate itself. `check_index_floor()` conflates *report once* with *enforce* (`src/pm/index_contract.cppm:110`), while its only caller uses the return value as the gate (`src/pm/package_fetcher.cppm:593`). So within one process the **first** lookup into a floor-violating index is blocked and every later one is silently let through.

`fetch_template_package()` looks up twice — `ns=""` then `ns="compat"`:

| lookup | candidates | result |
|---|---|---|
| `ns=""` | `imgui.lua`, `compat.imgui.lua` | blocked by floor (E0006 printed once) — `imgui.lua` never seen |
| `ns="compat"` | `compat.imgui.lua` | gate already "reported" → passes → Form B hit |

Reproduced with the exact binary CI installs (0.0.100) against the real index (floor 0.0.102):

```
$ .../xim-x-mcpp/0.0.100/mcpp new abc1 --template imgui
error: index requires mcpp >= 0.0.102 but this is mcpp 0.0.100 [E0006]
error: package 'imgui@1.92.8' has no mcpp.toml        # byte-identical to CI

$ MCPP_INDEX_FLOOR=ignore .../0.0.100/mcpp new abc2 --template imgui
     Created abc2 (template imgui@0.0.6:window)       # same binary, same index
```

Same binary, same index, only the gate differs — so candidate ordering and Form A/B are not the cause. The leak also defeats the floor contract everywhere else (`prepare.cppm` uses the same gate), which is exactly what mcpp-index #107 raised the floor to prevent. Introduced in 53b6ead (0.0.85, #200), i.e. present since the feature landed.

With this PR applied the same CI step still exits 1 — just with `template package 'imgui' not found in the index` instead. A separate PR fixes the gate and will close #265.

### On merging this

Fine to merge as hardening. Verified it cannot regress anything today: `imgui.lua` and `templates.lua` have no `mcpp` field at all (`Absent`), all 40 `TableBody` descriptors in the index are `compat.*` source packages, and a `TableBody` package could never have passed the `root/mcpp.toml` check anyway — so the skip only turns an unavoidable failure into a clearer message. It is also textually orthogonal to the gate fix.

Two things before merge:

1. Status checks have not run on this fork PR — please let Linux / macOS / Windows CI go green first (the last box in your test plan).
2. Consider a one-line comment at the `continue` stating *why* Form B can never provide templates, so the skip is not mistaken for the #265 fix later.

PR #300, opened 2026-07-28
fix: statically link build.mcpp to avoid missing DLL issues

### LINKED ISSUE #299: fix: build.mcpp binary fails at runtime with missing DLL on Windows
## Problem

On Windows, after mcpp compiles a user's `build.mcpp`, the resulting binary fails at runtime with exit code `-1073741511` (0xC0000135, `STATUS_DLL_NOT_FOUND`).

## Reproduction

- OS: Windows
- Toolchain: GCC 16.1.0 (MinGW)
- Steps:
  1. Run `mcpp build` on a project with a `build.mcpp`
  2. The compile succeeds, but the execution of `build.mcpp.exe` fails with exit code `-1073741511`

## Diagnosis

Running `Dumpbin /DEPENDENTS .\build.mcpp.exe` reveals the binary depends on:
- `libgcc_s_seh-1.dll`
- `libstdc++-6.dll`

These are dynamic libraries from the GCC toolchain. If the toolchain's `bin/` directory is not in `PATH`, the binary cannot find them at runtime.

Manually adding the GCC bin path to the environment variable `PATH` resolves the issue, but this is a workaround, not a fix.

## Root Cause

The `host_base_flags()` function in `src/build/build_program.cppm` compiles `build.mcpp` with dynamic linking by default. The produced binary dynamically links against the toolchain's runtime DLLs.

## Proposed Fix

Compile `build.mcpp` with `-static` (GCC: `-static -static-libgcc -static-libstdc++`; Clang: `-static`) so the binary is self-contained and does not depend on runtime DLLs. Also remove `-Wl,-rpath,` and `-Wl,--dynamic-linker=` flags which are unnecessary under static linking.

ZheFeng7110:
## Summary

- Statically link the compiled `build.mcpp` binary so users don't need to configure dynamic library paths
- Add `-static` flag to `host_base_flags()` for both Clang and GCC paths
- Remove `-Wl,-rpath,` and `-Wl,--dynamic-linker=` flags that are unnecessary under static linking

## Background

When mcpp compiles a user's `build.mcpp`, it produces a native binary that runs before the main build. Currently this binary is dynamically linked against the toolchain's runtime libraries (e.g. `libgcc_s_seh-1.dll`, `libstdc++-6.dll` on Windows; `libc++.so`, `libunwind.so` on Linux).

If the toolchain's library paths are not in the system's search path (`PATH` on Windows, `LD_LIBRARY_PATH` on Linux), the binary fails at runtime.

**Windows case**: exit code `-1073741511` (0xC0000135, `STATUS_DLL_NOT_FOUND`). `Dumpbin /DEPENDENTS` shows dependencies on `libgcc_s_seh-1.dll` and `libstdc++-6.dll`.

## Changes

- `src/build/build_program.cppm` — `host_base_flags()` function (lines 165-226):
  - Clang path: add `-static`, remove all `-Wl,-rpath,` and `-Wl,--dynamic-linker=`
  - GCC path: add `-static`, remove all `-Wl,-rpath,`
  - Keep `-L` flags (still needed to locate `.a` archives under static linking)

## Test plan

- [x] `mcpp build` passes (Windows, LLVM 20.1.7)
- [ ] E2E tests pass (Linux/macOS CI)
- [ ] `build.mcpp` binary runs without requiring PATH/LD_LIBRARY_PATH setup

Closes #299

### DIFF
--- a/src/build/build_program.cppm
+++ b/src/build/build_program.cppm
@@ -177,6 +177,7 @@ std::vector<std::string> host_base_flags(const mcpp::toolchain::Toolchain& tc) {
     // pipeline regenerates the cfg deterministically anyway.
     if (mcpp::toolchain::is_clang(tc)) {
         if constexpr (!mcpp::platform::is_linux) return f;
+        f.push_back("-static");
         const auto dm = mcpp::toolchain::resolve_clang_driver(tc);
         if (dm.hasCfg) {
             f.push_back("--no-default-config");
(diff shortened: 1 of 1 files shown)


Sunrisepeak:
- https://github.com/mcpp-community/mcpp/pull/298 

同类问题 合并到 298 一起合入

PR #307, opened 2026-07-29
feat: reject mcpp add for packages not present in the index

### LINKED ISSUE #305: 希望mcpp add 包指令增加检查
现在的mcpp add 包指令是直接修改toml配置文件，但是没有检查包是否存在，mcpp add 114514@114514这种明显没意义的包也能直接加，build的时候才会检查，希望提前一些，add的时候去仓库简单检索一下包是否存在

ZheFeng7110:
## Summary
- Validate that the requested package exists in a configured index before mutating `mcpp.toml`.
- If the package is not found, print an error and exit without writing the dependency.
- Propagate `config::load_or_init` failures instead of silently skipping the check.
- Update the `tests/e2e/12_add_command.sh` E2E test to use real packages and add a regression case for the missing-package scenario.

Closes #305

## Test plan
- [x] `mcpp build` passes
- [x] `tests/e2e/12_add_command.sh` passes

### DIFF
--- a/src/build/prepare.cppm
+++ b/src/build/prepare.cppm
@@ -39,6 +39,7 @@ import mcpp.fetcher;
 import mcpp.fetcher.progress;
 import mcpp.pm.resolver;
 import mcpp.pm.index_spec;
+import mcpp.pm.index_route;
 import mcpp.pm.mangle;
 import mcpp.pm.compat;
 import mcpp.pm.dep_spec;
@@ -694,20 +695,10 @@ prepare_build(bool print_fingerprint,
                     m->targetOverrides[triple] = entry;
                 }
             }
-            // Inherit workspace indices if member doesn't define any. A
-            // relative `[indices].path` was declared at the WORKSPACE root,
-            // so it must resolve against `*root` (still the workspace root
-            // here), not the member directory — otherwise every member
-            // needs its own `../`-prefixed copy of the same declaration
-            // (#224).
-            if (m->indices.empty() && !wsManifest->indices.empty()) {
-                m->indices = wsManifest->indices;
-                for (auto& [_, idx] : m->indices) {
-                    if (idx.is_local() && idx.path.is_relative()) {
-                        idx.path = std::filesystem::weakly_canonical(*root / idx.path);
-                    }
-                }
-            }
+            // Inherit workspace indices if member doesn't define any. `*root`
+            // is still the workspace root here, which is what a relative
+            // `[indices].path` was written against (#224).
+            mcpp::project::inherit_workspace_indices(*m, *wsManifest, *root);
 
             mcpp::ui::status("Workspace", std::format("building member '{}'", targetMember));
             root = memberDir;
@@ -730,14 +721,7 @@ prepare_build(bool print_fingerprint,
                     }
                 }
                 // Inherit workspace indices if member doesn't define any
-                if (m->indices.empty() && !wsm->indices.empty()) {
-                    m->indices = wsm->indices;
-                    for (auto& [_, idx] : m->indices) {
-                        if (idx.is_local() && idx.path.is_relative()) {
-                            idx.path = std::filesystem::weakly_canonical(wsRoot / idx.path);
-                        }
-                    }
-                }
+                mcpp::project::inherit_workspace_indices(*m, *wsm, wsRoot);
             }
         }
     }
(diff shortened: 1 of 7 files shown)


speak-agent (commented):
### Code review

Right problem (#305), but the gate uses the narrowest lookup path in the codebase, so two classes of valid packages are now rejected.

**1. Dotted selectors are rejected.** `cmd_add` passes the whole selector as `shortName`, but `read_xpkg_lua`'s identity gate requires `id.name == shortName` and SPEC-001 makes `package.name` a single atomic segment. `capi.lua` is a real index entry (`namespace = "mcpplibs.capi", name = "lua"`) and a documented form; verified through the same lookup: `mcpp new probe --template capi.lua` → `template package 'capi.lua' not found in the index`. The PR also deletes the e2e case that covered this (added by 37cbc83) and contradicts the comment directly above it.

https://github.com/mcpp-community/mcpp/blob/5032fe6ed87eb069aa8209d1e3d402cfa5e8f6e5/src/pm/commands.cppm#L83-L88
https://github.com/mcpp-community/mcpp/blob/fde3b7005750d096e663df95ddec23a6ac932d90/docs/05-mcpp-toml.md#L305-L311

**2. `[indices]` custom/local indices are ignored.** Build-time lookup is a three-way dispatch (local path index / project data / global); the gate only uses the global one and never reads `[indices]` from `mcpp.toml`. In the `120_ws_root_indices.sh` layout, `mcpp add x:widget2@1.0.0` fails while `mcpp build` resolves it.

https://github.com/mcpp-community/mcpp/blob/fde3b7005750d096e663df95ddec23a6ac932d90/src/build/prepare.cppm#L1513-L1526

**3. No index refresh before the lookup.** `mcpp build` and `mcpp index search` call `ensure_index_fresh` first; the gate only calls `load_or_init`. A stale or newly published entry yields a false "not found", and the message gives no way out — compare `create.cppm`, which says "check the name, or run `mcpp index update`". The message also prints `shortName` only, dropping the namespace (`compat:nope` → `package 'nope' not found`).

https://github.com/mcpp-community/mcpp/blob/fde3b7005750d096e663df95ddec23a6ac932d90/src/build/prepare.cppm#L1320-L1324
https://github.com/mcpp-community/mcpp/blob/fde3b7005750d096e663df95ddec23a6ac932d90/src/scaffold/create.cppm#L55-L61

**4. e2e 12 now requires a full sandbox bootstrap but declares no `# requires:`.** It exports an isolated `MCPP_HOME`, so every run downloads xlings, fetches the package index, patchelf and ninja. `fresh-sandbox` is deliberately not enabled on Windows, and without a requires line this test is not skipped there.

https://github.com/mcpp-community/mcpp/blob/5032fe6ed87eb069aa8209d1e3d402cfa5e8f6e5/tests/e2e/12_add_command.sh#L10-L21
https://github.com/mcpp-community/mcpp/blob/fde3b7005750d096e663df95ddec23a6ac932d90/tests/e2e/run_all.sh#L91-L94

**Suggested shape:** resolve the input with `resolve_dependency_selector`, then check the candidates through the same three-way dispatch `prepare.cppm` uses (`readStrictLuaForCandidate`), accepting on any candidate hit; call `ensure_index_fresh` first; put the full spec plus an `mcpp index update` hint in the error; restore the dotted-selector e2e case and add `# requires: fresh-sandbox` to test 12.

speak-agent:
Pushed a follow-up commit reworking the gate. The direction (#305) was right; the lookup it was built on was the narrowest one in the tree, so it refused packages that resolve fine at build time.

**Routing is now shared.** `mcpp.pm.index_route` owns "which index answers for a namespace, and how its descriptors are read", and both `mcpp add` and `mcpp.build.prepare` go through it. That closes the two false rejections:

- dotted selectors — `capi.lua` is a namespace path meaning `(mcpplibs.capi, lua)`, then `(capi, lua)`. Nothing in any index is *named* `capi.lua` (`package.name` is a single atomic segment, SPEC-001 §3.2), so a literal probe could never match. mcpp's own `mcpp.toml` is written in that form.
- anything served by a project `[indices]` entry — the old probe read only the global registry, while dependency resolution dispatches across local-path, project-clone and global transports.

`[indices]` workspace inheritance moved into `mcpp.project` for the same reason: two copies of that rule is how the two readers ended up disagreeing.

**Refuse only what is provably absent.** A lazily-cloned git index, or a namespace no readable index covers (xim descriptors declare no `namespace` at all, so `(xim, nasm)` can never satisfy the identity gate), now reports *unverified* instead of failing — the same fall-through `prepare` already applies to lazy git indices.

**Cost and diagnostics.** A stale registry is refreshed once before refusing, and only when the registry is the index that would have answered, so a package already on disk costs zero network round-trips. Errors list the identities tried plus a cross-namespace did-you-mean, matching prepare's resolution error. A version the descriptor doesn't publish for this platform is a warning with the available list, not a failure — version tables are per-OS, so "absent here" is not "absent".

**Tests.** `tests/e2e/12_add_command.sh` serves every package from a local `[indices] path`, so it covers the dotted, project-index and workspace-inheritance routes offline instead of depending on the shared registry's state; it also drops its isolated `MCPP_HOME`, which would otherwise force a full sandbox bootstrap (xlings, index fetch, patchelf, ninja) on a test that is otherwise local file manipulation. New `tests/unit/test_pm_index_route.cpp` pins the routing rules, including "a literal dotted short name never matches".

Out of scope here, worth a separate issue: `mcpp::pm::resolve_semver` also reads the global registry directly, so a SemVer constraint on a package served by a project `[indices]` entry cannot resolve. Same family of defect, different code path.

PR #330, opened 2026-08-01
feat(pm): make mcpp.lock authoritative for git deps (2026.8.1.2)

### LINKED ISSUE #329: feat: use mcpp.lock commit as offline anchor for git branch dependencies
## 问题

`[dependencies] ccc = { git = "https://github.com/ZheFeng7110/ccc", branch = "develop" }` 这种 branch 引用的 git 依赖，每次 `mcpp build` 都会执行 `git ls-remote`，即使 `mcpp.lock` 已经锁定了具体 commit。

当前 `src/build/prepare.cppm` 对 branch 依赖的处理是：每次 build 都 `ls-remote` 分支 HEAD，然后解析为 commit 再作为缓存 key。这导致：
- 慢网络下频繁等待
- 离线/无网环境下构建失败
- 与 `mcpp.lock` 的“锁定语义”不一致

## 期望行为

把 `mcpp.lock` 作为 git 依赖的离线锚点：
- 若 lock 已记录该 git 依赖的 commit，且本地 `~/.mcpp/git/<hash>` 缓存的 HEAD 与 lock commit 一致，则直接复用缓存，不再 `ls-remote`。
- 仅当 lock 不存在/无该包/无 commit/缓存 commit 不匹配时，才执行网络请求。
- `mcpp update` 清除 lock 条目后，下次 build 重新 `ls-remote` 并写入新 commit。
- 对 `rev` 和 `tag` 也纳入 lock 决策，保持统一语义。

## 离线模式

- `--offline` 下，若 lock 有 commit 且缓存命中，应允许构建。
- 若 lock 有 commit 但缓存缺失，离线模式应失败（无法 clone）。
- 若 lock 无 commit，离线模式应失败，提示需要先用非离线模式或 `mcpp update`。

## 涉及模块

- `src/pm/lock_io.cppm`：新增 lock 中 git source 的解析工具
- `src/build/prepare.cppm`：读取 lock、命中 branch 缓存、离线模式
- `tests/e2e/24_git_dependency.sh`：增加 branch 锁定/不重新 ls-remote 的用例

ZheFeng7110:
## Summary

Makes `mcpp.lock` a resolution input for git dependencies — and an authoritative
one. Before this change the lock was write-only: a branch-based git dependency
(`branch = "develop"`) re-ran `git ls-remote` every time `prepare_build` ran, and
the commit written into `mcpp.lock` was never read back by anything.

After this change the lock is read first. A branch that already has a commit
recorded resolves from it with no network round-trip, and **that commit is what
gets built whether or not the local clone survived** — evicting `~/.mcpp/git`, or
cloning the project on another machine, can no longer put the build on a newer
branch tip. `mcpp update <dep>` drops the entry and remains the one way a branch
advances.

Closes #329.

## Why authoritative, and not "an anchor the cache confirms"

A lock that only counts while the clone happens to exist hands the "which commit
do we build" decision to the survival of a cache directory. `mcpp new` does not
gitignore `mcpp.lock` — it is meant to be committed — so the failure is ordinary:
a teammate clones the project, has no `~/.mcpp/git`, and silently builds a
different commit while rewriting the lock.

This is also the dominant convention: `cargo build` on a git branch dependency
builds the locked revision and `cargo update` re-resolves it — the same split
mcpp already follows for profile defaults.

Measured, with the branch moved to v2 and the cache removed (this is the new
assertion in `tests/e2e/24_git_dependency.sh`):

| | build output |
|---|---|
| lock as cache hint | `branch dep v2` — silently advanced, lock rewritten |
| lock authoritative | `branch dep v1` — the commit the lock recorded |

## Shape

The block collapses into two independent questions, each with at most one network
operation and therefore exactly one `--offline` gate:

1. **Which commit?** `tag`/`rev` name one outright. A `branch` is floating:
   `mcpp.lock` answers it, else `git ls-remote`.
2. **Is it on disk?** The commit selects the cache directory; a miss is a clone.

The separate tag/rev leg is gone with it — along with its per-build `from cache`
noise line and its `"is locked but its local cache is missing"` message, which
was emitted for dependencies that were never in the lock.

## Also fixed here

- **`--offline` no longer refuses a local git remote.** `docs/05-mcpp-toml.md`
  defines `--offline` as "never touch the network … anything already installed
  still builds", and the dependency-download gate in `prepare.cppm` draws the
  same line in a comment. A `git =` naming a local directory (or a `file://`
  URL) is served by filesystem reads, so refusing it buys no isolation. Remotes
  are classified by shape, keeping a Windows drive letter (`C:\repo` — colon,
  no `@`) on the local side.
- **A clone killed between `git clone` and `git checkout` no longer serves the
  wrong commit forever.** The cache directory is named after the commit but was
  only checked for existence. Branch deps now compare `git rev-parse HEAD` and
  re-clone on mismatch (tag/rev keep their ref name as the identity, so there is
  nothing to compare).
- **Windows: the clone no longer depends on `cd` changing drive.** `git clone …
  && cd <dir> && git checkout` does not change drive in cmd.exe without `/d`,
  and the cache root routinely lives on a different one than the project. Every
  other cross-drive site in the repo (`process.cppm`, `msvc.cppm`) writes
  `cd /d`; this one did not. Now `git -C <dir>`, which needs no shell at all.
- **An unreadable `mcpp.lock` is reported as degraded, not as a plain warning.**
  Per `src/diag.cppm` that channel is for "the engine did less than asked" and
  requires an impact — here every git branch dep silently falls back to the
  network and may move past the recorded commit. As a warning it was invisible
  to `--strict`.
- **`parse_git_source` no longer splits a tag/rev whose name contains `@`.**
  Only `branch` entries carry `@<commit>` (that is what the writer emits), so a
  tag legitimately named `v1.0@rc1` was being read as ref `v1.0` + commit `rc1`.
  Branch entries still split on the last `@`, so `feat@v2` round-trips.

## Note on the original framing

The first revision described this as removing an `ls-remote` from "every
`mcpp build`". That overstates it: a no-op rebuild takes the fast path in
`cmd_build.cppm`, which returns before `prepare_build` runs, so no `ls-remote`
happened there either. The cost this removes is on builds that *do* re-prepare —
sources changed, `mcpp.toml` touched, a `--profile`/`--target`/`--features` flag
passed, or after `mcpp clean`. That is also why the e2e has to `mcpp clean`
first: without it the assertion passes on `main` too.

## Changes

- `src/pm/lock_io.cppm`: `LockedGitSource` + `parse_git_source()`.
- `src/build/prepare.cppm`: load `mcpp.lock` up front; two-phase git resolution;
  local-remote classification; cache integrity check; `git -C`.
- `src/pm/commands.cppm`: correct the `cmd_update` comment — dropping a lock
  entry is now the thing that lets a branch advance, not a no-op.
- `docs/05-mcpp-toml.md`, `docs/zh/05-mcpp-toml.md`: git deps + `mcpp.lock`, and
  the `--offline` row.
- `CHANGELOG.md`, `mcpp.toml`, `src/toolchain/fingerprint.cppm`: 2026.8.1.2.
- `tests/e2e/24_git_dependency.sh`: offline rebuild from the lock; the branch
  does not advance when the cache is evicted.
- `tests/unit/test_pm_lock_io.cpp`: `@` in branch names, `@` in tag/rev names,
  empty commit, scp-like URLs.

## Test plan

- [x] `mcpp build`
- [x] `mcpp test` (46 passed, 0 failed)
- [x] `tests/e2e/24_git_dependency.sh`
- [x] A/B: the new e2e fails on the previous head (builds v2) and passes here
- [x] `.github/tools/check_version_pins.sh`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,28 @@
 > 本文件追踪 `mcpp-community/mcpp` 公开仓的版本演进。
 > 格式参考 [Keep a Changelog](https://keepachangelog.com/zh-CN/1.1.0/)。
 
+## [2026.8.1.2] — 2026-08-01
+
+### 新增
+
+- **`mcpp.lock` 成为 git 依赖的解析输入,而且是权威的那一份。** 在此之前 lock 只写不读:`branch = "develop"` 这类依赖每次进 `prepare_build` 都要打一次 `git ls-remote`,而写进 lock 的那个 commit 从来没有人看过。现在**先读 lock**——分支已经解析过就直接用记录的 commit,不再有网络往返。
+
+  「权威」是这里的关键词,不是「缓存的提示」。一个只在本地克隆恰好还在时才生效的 lock,等于把「构建哪个 commit」这个决定交给了缓存目录的存活状况:清一次 `~/.mcpp/git`、或者换一台机器 clone 同一个工程,构建就会静默滑到分支的新头上,并把 lock 改写成新值——**而 `mcpp new` 生成的 `.gitignore` 并不忽略 `mcpp.lock`,它本来就是要提交的**。实测(见 e2e 24 新增断言):分支移到 v2、缓存删除后,旧行为构建出 v2,新行为构建出 lock 里记的 v1。要新的分支头依然只有一条路——`mcpp update <dep>`,它丢掉 lock 条目,下次构建重新解析。
+
+  连带把这块的结构收敛成两个各自独立的问题,各带**恰好一个** `--offline` 闸:**(1) 是哪个 commit** —— `tag`/`rev` 自带答案,`branch` 由 lock 回答、否则 `ls-remote`;**(2) 它在不在盘上** —— commit 选定缓存目录,miss 就 clone。原来 tag/rev 有一条自己的分支腿、打一行每次构建都出现的 `from cache` 噪声,并且在 dep 根本没进 lock 时报「is locked but its local cache is missing」;这条腿现在整个不需要了。
+
+### 修复
+
+- **`--offline` 不再拒绝本地 git 远端。** `docs/05-mcpp-toml.md` 写明 `--offline` 的语义是「完全不碰网络……已安装的东西照常构建」,`prepare.cppm` 里依赖下载闸的注释也把线画在同一处:「这条线以上全是本地操作,一个依赖齐备的离线构建必须成功」。但 `git = "../sibling-repo"` 这种指向本地目录的远端,`ls-remote`/`clone` 都只是文件系统读取,拒绝它买不到任何隔离性。现在按远端形态判定——`file://`、以及不带 scheme 也不是 `git@host:path` 的存在路径,算本地(Windows 盘符 `C:\repo` 含冒号但不含 `@`,因此仍归本地)。
+
+- **克隆被中途杀掉后不再永久提供错误的 commit。** 缓存目录以 commit 命名,但内容是 `git clone` 之后再 `git checkout` 两步做出来的;进程死在两步之间,目录名和 HEAD 就对不上了,而后续构建只检查目录存不存在。现在分支依赖会比对 `git rev-parse HEAD`,不符即删除重克隆(tag/rev 以 ref 名为身份,无可比之物)。
+
+- **git 克隆在 Windows 上不再依赖 `cd` 跨盘。** 克隆命令用的是 `git clone … && cd <dir> && git checkout`,而 cmd.exe 的 `cd` 不带 `/d` 是不换盘的——缓存根(`MCPP_HOME`)与工程不同盘是常态。仓库里其它跨盘位置(`process.cppm`、`msvc.cppm`)都写了 `cd /d`,这里漏了。改用 `git -C <dir>`,连 shell 都不需要。
+
+- **`mcpp.lock` 读取失败改用 degraded 通道。** 按 `src/diag.cppm` 的划分,「引擎做得比要求的少」属于 Degraded 且必须给出 impact——lock 读不出来,每个 git 分支依赖都会退回网络解析、并可能越过记录的 commit。原来记的是普通 warning,`--strict` 提升不到它。
+
+- **`parse_git_source` 不再切开名字里带 `@` 的 tag/rev。** 写侧只对 `branch` 追加 `@<commit>`,解析侧却无条件按最后一个 `@` 切分,于是一个叫 `v1.0@rc1` 的 tag 会被读成 ref `v1.0` + commit `rc1`。现在只有 branch 走这条切分(且仍按最后一个 `@`,以容纳 `feat@v2` 这类分支名)。
+
 ## [2026.8.1.1] — 2026-08-01
 
 ### 修复
--- a/docs/05-mcpp-toml.md
+++ b/docs/05-mcpp-toml.md
@@ -339,9 +339,10 @@ glfw = "3.4"                    # Explicit namespace, skips the mcpplibs-first c
 [dependencies]
 mylib = { path = "../mylib" }
 
-# Git dependency
+# Git dependency — pick exactly one of tag / branch / rev
 [dependencies]
 mylib = { git = "https://github.com/user/mylib.git", tag = "v1.0.0" }
+applib = { git = "https://github.com/user/applib.git", branch = "develop" }
 
 # Long-form dep spec: features and backend knobs
 [dependencies]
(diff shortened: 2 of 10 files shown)


speak-agent:
### Code review

The main path is correct: `computeGitRoot()` matches the real cache-key formula field for field, so a lock hit provably lands on the existing clone and skips it; `readCacheHead` compares against a detached HEAD left by `git clone --depth 1 --branch X && git checkout <commit>`, so both sides are full 40-char shas; and the skip path still repopulates `root_git_lock_identities`, so the lock stays idempotent. `mcpp update <dep>` erases by `p.name`, which is the same key the anchor lookup uses, so the refresh path in the e2e still holds.

Found 4 issues:

1. The new e2e assertion is vacuous — it passes identically on `main` without this change. `ls-remote` appears in build output only inside the failure message `"git ls-remote of '{}' failed"`; on the success path the command is captured with `2>&1` into `r.output` and never printed, so the grep can never match whether or not the network call happened. And the only case that would match is a failed build, which `set -e` (line 5) already aborts on at the `build2=$(...)` assignment. The `Cloning` guard doesn't discriminate either: on `main` the second build resolves to the same commit, hence the same `gitRoot` hash, hence no re-clone. Net effect: the headline behaviour of this PR has no test coverage, and the e2e box in the description is unchecked. The positive signal already exists — this PR adds `ui::info("Resolved", "... from lock")`, so `grep -q 'from lock'` would actually assert it.

https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/tests/e2e/24_git_dependency.sh#L155-L160

2. The cache-key formula is now derived in two places. `computeGitRoot` and the block that actually creates the directory are byte-identical today, so this is not a live bug — but any future axis added to one side (profile, triple) makes the anchor miss forever, silently degrading back to per-build `ls-remote` with no error and, given issue 1, no failing test. The second site should just call `computeGitRoot(resolvedGitRev)`.

https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/src/build/prepare.cppm#L3027-L3035

https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/src/build/prepare.cppm#L3109-L3117

3. `LockedGitSource::url` is parsed but never compared — the anchor matches on `refKind` + `ref` only. Online this self-heals (a changed URL changes the `gitRoot` hash, the cache misses, and it falls through to `ls-remote`), but under `--offline` a repo URL change is reported as `locked to commit <old-repo-commit> but its local cache is missing or stale`, pointing the user at a stale cache when the actual cause is a different repository. Adding `anchor.url == spec.git` to the match would fix both the diagnostic and the fallback.

https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/src/build/prepare.cppm#L3018-L3027

4. Both offline diagnostics recommend `mcpp update <dep>`, which cannot help offline. For git deps that command only drops the lock entry (`src/pm/commands.cppm:449-458`) — it never resolves or clones. In the "lock has no commit, cannot resolve offline" case, following that advice deletes the only remaining anchor. `run without --offline` is the only real remedy and should be the sole suggestion here.

https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/src/build/prepare.cppm#L3061-L3074

Two smaller notes, same PR:

- The comment in `cmd_update` explaining why that command used to be a no-op rests on "the build path never reads mcpp.lock (prepare writes it and nothing on that path loads it)". This PR makes that false, and it is load-bearing — the next person changing `mcpp update` will reason from it. Worth updating in the same change.

  https://github.com/mcpp-community/mcpp/blob/7f1489d24c88cd849d9cc0cdd7d5e9b3d77dc31a/src/pm/commands.cppm#L413-L419

- `parse_git_source` splits on the first `@`, so a branch name containing `@` (`feat@v2`) parses wrong. The consequence is only a safe degradation (anchor misses, falls back to `ls-remote`), but `rfind` matches the write format better. Relatedly, `branch=develop@` yields `optional("")`, which passes the `if (lockedCommit)` test in the offline branch and prints `locked to commit ` with an empty sha — the wrong one of the two messages.

  https://github.com/mcpp-community/mcpp/blob/c3e709543dfedd272c7d4a78acabdf7d80280125/src/pm/lock_io.cppm#L198-L206

Behaviour note, not a defect: `ui::info("Resolved", ...)` now prints on every build for both branch and tag/rev deps, including no-op incremental builds that used to be silent. The tag/rev one is purely informational — it gates on `exists()` and changes no decision — so it may belong behind `--verbose`.

PR #343, opened 2026-08-03
ci: workspace mcpp bootstrap pin -> 2026.8.3.3 (released + indexed)

Sunrisepeak:
Routine follow-up to #342 / v2026.8.3.3.

The bootstrap pin only moves once the release it names exists and is reachable
through the index. Both hold:

- v2026.8.3.3 published on linux x86_64 / linux aarch64 / macOS ARM64 / Windows x64
- mirrored to `xlings-res/mcpp` on GitHub **and** GitCode — the GitCode copy verified
  by a real GET whose sha256 matches the published one (the release workflow's mirror
  step hit the 10-minute action timeout *after* completing all four uploads, so the
  index bump was finished by hand)
- `xim-pkgindex` tracks it as latest
- end to end: `xlings install mcpp@2026.8.3.3` → `xlings use` → `mcpp --version` reports
  2026.8.3.3

And the released binary reproduces the B3 fix on a fresh project:

```
ninja declares: bin/finalprobe.exe
declared output EXISTS ✓
no relink on rebuild ✓
```


Sunrisepeak:
Closing — this should not have been opened.

memory/release-bootstrap-pin-two-groups.md and
.agents/docs/2026-07-30-issue315-implementation-plan.md:132 both say the
bootstrap pin does NOT move with a release: it is the version BOOTSTRAPPED FROM,
any released mcpp that can build the current tree, and lagging a few versions is
the normal state. check_version_pins.sh only enforces an upper bound (not newer
than what is being built), not equality.

I turned 'may move once the release exists and is indexed' into 'should move
every release', and called it a convention. It is not one. The pin stays at
2026.8.3.2, which is released, indexed, and builds this tree fine.

PR #438, opened 2026-08-16
test(e2e): 239 could pass or skip, never fail

Sunrisepeak:
## Summary

The skip was decided **after** the install, by pattern-matching the failure text, and one of the patterns was `*"index"*`. Nearly every mcpp command prints "package index" somewhere, so **every genuine install failure took the skip branch**.

It hid a real one, on the first Windows run where the package was actually installable:

```
tar: Cannot connect to C: resolve failed
tar -xf "C:\Users\...\.payloads\Microsoft.VC...vsix" -C "..."
[error] msvc installed but registered none of the programs it declares
```

The install ran for **135 seconds**, failed, and this script printed `PASS`.

## Fix

Skip is decided **before** the work, by a positive check for the thing whose absence would make the test impossible (no msvc row in `toolchain list`). Everything after that is a failure, and the output is printed rather than folded into a one-line message.

> A skip decided by the shape of a failure is not a skip; it is a way of not looking.

The recipe-side fix for what it hid is openxlings/xim-pkgindex#632 (GNU tar reads `C:` as a hostname; bsdtar does not, which is why the index's own `windows-test` passed).

## Test plan

- [x] `bash -n` clean
- [ ] `ci-windows-e2e` — 239 now either runs for real or skips for a stated, checkable reason

### DIFF
--- a/tests/e2e/239_msvc_managed_toolset.sh
+++ b/tests/e2e/239_msvc_managed_toolset.sh
@@ -36,16 +36,30 @@ restore() {
 trap restore EXIT
 cd "$TMP"
 
-# 1) install it. This runs BEFORE the discoverability check below, so that a
-#    runner which cannot reach the index skips instead of failing an assertion
-#    about a list the index would have filled in.
+# 0) Decide SKIP here, from a positive check, and never again.
+#
+#    ⚠️ This used to be decided AFTER the install by pattern-matching the
+#    failure text, and one of the patterns was `*"index"*`. Nearly every mcpp
+#    command prints "package index" somewhere, so every genuine install
+#    failure took the skip branch: the test could pass or skip, never fail.
+#
+#    It hid a real one. `tar -xf "C:\...vsix"` fails under GNU tar, which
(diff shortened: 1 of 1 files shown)


Sunrisepeak:
Folded into #440 and closing there.

#440 has to re-run its Windows e2e anyway (it was blocked on openxlings/xim-pkgindex#637 / #639), so carrying this as a separate PR would have cost a second full CI cycle for a change to the same file — #440 already touches `239_msvc_managed_toolset.sh`.

The commit is preserved verbatim, plus two follow-ups that today's run argued for:

- **the skip asks for `$TOOLSET`, not the family.** Grepping for "msvc" is true during an index publish window — the family is listed, this version is not yet — and now that a failure is hard, that would report a timing artifact as a defect.
- **the install must report a Windows SDK.** `install` prints `windows sdk: <version> (<root>)`, but the script discarded its output on success. That is exactly why a half-installed SDK said nothing here and surfaced ~100 lines later as `LNK1104: cannot open file 'kernel32.lib'` with no line naming the SDK.

PR #441, opened 2026-08-16
feat(toolchain): `gcc@system` is rejected, and says what to do instead

Sunrisepeak:
`@system` is available for **MSVC only**, and that is deliberate rather than incidental.

xlings depends on the host as little as it can: a toolchain comes from a payload, which is what makes *"the manifest says 14.44.35207"* true on **every** machine instead of on the one that happened to have it. Offering `gcc@system` invites that uncertainty back, and the alternative is one command away.

MSVC is the exception because **Windows** is: Visual Studio is frequently already installed and cannot always be redistributed, so refusing to use it would mean refusing to build. That is a platform fact, not an instance of a general capability — so it is not generalised.

### Why reject rather than leave unimplemented

Until now `gcc@system` parsed into a spec that meant "no version", then failed somewhere further in with a message about something else. It is now refused at the parse, with the version form and the reason for the one exception:

```
'gcc@system' is not a thing — mcpp does not build with the machine's own gcc.
  A toolchain comes from a payload, so a manifest means the same thing on every machine.
  Name a version instead: `gcc@<version>` (`mcpp toolchain list` shows what is available).
  Only `msvc@system` exists, because Visual Studio cannot always be redistributed.
```

The test asserts the refusal **and** that the message names the alternative — a refusal that does not is just a wall. Verified against a guard that never fires.

> Follows the architecture review (`.agents/docs/2026-08-16-toolchain-architecture-review.md` §1). That section's first draft proposed generalising `@system` to all families; it was wrong, and the doc now records both the correction and the reason.

### DIFF
--- a/src/toolchain/registry.cppm
+++ b/src/toolchain/registry.cppm
@@ -229,6 +229,35 @@ parse_toolchain_spec(std::string compilerArg,
             "supported alias like mingw / musl-gcc)", compilerArg));
     }
 
+    // `@system` means "whatever this machine has", and it is deliberately
+    // available for MSVC ONLY.
+    //
+    // xlings depends on the host as little as it can: a toolchain comes from
+    // a payload, which is what makes "the manifest says 14.44.35207" true on
+    // every machine instead of on the one that happened to have it. Offering
+    // `gcc@system` would invite the uncertainty back in, and the alternative
+    // is one command away.
+    //
+    // MSVC is the exception because Windows is: Visual Studio is frequently
+    // already installed and cannot always be redistributed, so refusing to
+    // use it would mean refusing to build. That is a platform fact, not a
+    // general capability, so it is not generalised.
+    //
+    // Rejected here rather than left unimplemented: an unimplemented spelling
+    // fails somewhere further in with a message about something else.
+    if (norm->version == "system" && norm->family != "msvc") {
+        return std::unexpected(std::format(
+            "'{}@system' is not a thing — mcpp does not build with the "
+            "machine's own {}.\n"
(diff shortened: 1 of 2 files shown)


Sunrisepeak:
Closing — this solves a problem I invented.

`gcc@system` never existed as a feature or a request. It came from the first draft of my own architecture review, where I read `is_system_toolchain()` being gated on `Family::Msvc` as an *asymmetry to remove* and proposed generalising it. That was backwards: xlings is a user-space OS and mcpp is built to minimise host dependence, so a host-provided toolchain is the thing the design avoids, not a capability it lacks. `msvc@system` exists because Visual Studio cannot always be redistributed — a platform concession, not a pattern.

Having been corrected on the direction, I then wrote a *rejection* for the spelling I had invented. That is scope created out of my own mistake, and the code is not free: it is a special case in the parser for a spelling nobody writes, which already fails today (it resolves to `xim:gcc@system` and is not found).

It also asserts something false. The message says "Only `msvc@system` exists", but `prepare.cppm:1376` has a deliberate bare-`system` escape hatch:

```cpp
} else if (tcSpec.has_value() && *tcSpec == "system") {
    // Explicit user opt-in to system PATH compiler — kept as escape hatch.
```

So the one place host dependence is genuinely offered is spelled `system`, with no family — and I was about to ship a message denying it.

The review doc's §1 is being corrected to drop the invented framing entirely. What survives there is the finding that was real and independent of it: one platform exception spread across 26 branches, three duplicated spellings of the same layout rule, and a resolution comment claiming 4 steps for a chain with 9 inputs.

PR #485, opened 2026-08-22
fix: include bound glibc in compiler environment

### LINKED ISSUE #484: Linux: private glibc ld.so.cache is not relocated, so gcc probe fails with /etc/ld.so.preload
## Summary

On a Linux host with `/etc/ld.so.preload`, the managed GCC 16.1.0 toolchain can fail before compilation starts. `mcpp build` fails while probing `g++ --version` with exit status 127.

The private glibc 2.44 loader cache contains absolute paths from the package build machine rather than the installed mcpp home. A preloaded host library whose indirect dependency is not covered by the executable's `DT_RUNPATH` therefore cannot locate `libdl.so.2`.

## Environment

- mcpp: `2026.8.21.3`
- managed compiler: `gcc@16.1.0`
- private libc: `xim-x-glibc@2.44`
- host has `/etc/ld.so.preload` containing `/lib/x86_64-linux-gnu/libonion.so`
- `libonion.so` has `DT_NEEDED: libdl.so.2` and no `RPATH`/`RUNPATH`

## Reproduction

```bash
MCPP_HOME=${MCPP_HOME:-$HOME/.mcpp}
GCC=$MCPP_HOME/registry/data/xpkgs/xim-x-gcc/16.1.0
GCC_RT=$MCPP_HOME/registry/data/xpkgs/xim-x-gcc-runtime/15.1.0

env -i PATH=/usr/bin:/bin LD_LIBRARY_PATH="$GCC_RT/lib64" \
  "$GCC/bin/g++" --version
```

Actual result:

```text
.../bin/g++: error while loading shared libraries: libdl.so.2: cannot open shared object file: No such file or directory
```

`mcpp build -v` then reports the probe command as exited with status 127, before `build.mcpp` or any project source is compiled.

The failure is independent of inherited `LD_LIBRARY_PATH` and `LD_PRELOAD`: it still reproduces with `env -i`. `/etc/ld.so.preload` is processed by the private loader.

## Evidence

The installed private cache is stale:

```bash
strings "$MCPP_HOME/registry/data/xpkgs/xim-x-glibc/2.44/etc/ld.so.cache" | grep libdl.so.2
# /home/xlings/.xlings_data/xim/xpkgs/fromsource-x-glibc/2.44/lib/libdl.so.2
```

The file actually exists at:

```text
$MCPP_HOME/registry/data/xpkgs/xim-x-glibc/2.44/lib/libdl.so.2
```

`LD_DEBUG=libs,files` identifies the complete failing edge:

```text
file=libonion.so [0]; needed by .../xim-x-gcc/16.1.0/bin/g++ [0]
file=libdl.so.2 [0]; needed by /lib/x86_64-linux-gnu/libonion.so [0]
... search path=$GCC_RT/lib64 (LD_LIBRARY_PATH)
... search cache=$MCPP_HOME/registry/data/xpkgs/xim-x-glibc/2.44/etc/ld.so.cache
... trying file=/home/xlings/.xlings_data/.../lib/libdl.so.2
```

The GCC executable itself has a correct `DT_RUNPATH` for the current private glibc directory, but it is `DT_RUNPATH`, so it is not transitive to the preloaded library's dependency.

As a confirmation, explicitly including the private glibc directory makes the exact command succeed:

```bash
GLIBC=$MCPP_HOME/registry/data/xpkgs/xim-x-glibc/2.44
env -i PATH=/usr/bin:/bin LD_LIBRARY_PATH="$GLIBC/lib:$GCC_RT/lib64" \
  "$GCC/bin/g++" --version
# g++ (xlings-fromsource) 16.1.0
```

## Expected behavior

A managed GCC toolchain should remain executable on a host that uses `/etc/ld.so.preload`; at minimum, mcpp should fail with a diagnostic identifying the broken private glibc cache rather than a generic compiler exit status.

## Possible fixes

1. Package/install fix: regenerate or relocate the private glibc `etc/ld.so.cache` after installation, or avoid shipping a cache with build-machine absolute paths.
2. Probe hardening: derive the compiler's bound glibc library directory and include it in the compiler child environment, rather than injecting only `gcc-runtime`.
3. Add a regression fixture using a preloaded library with an indirect dependency outside the compiler executable's direct dependency closure.

The immediate local workaround is rebuilding the private cache with its bundled `ldconfig`, but this should not be necessary for a managed toolchain.

ZheFeng7110:
## Summary
- add the exact `glibc@<version>` payload library directory to Linux compiler child environments
- keep runtime selection explicit; do not scan installed glibc versions
- cover the initial GCC `--version` probe failing with exit 127 when the directory is absent

Closes #484

## Test plan
- [x] `git diff --check`
- [x] managed GCC 16.1.0 `--version` succeeds with the bound private glibc plus gcc-runtime directories
- [x] managed GCC 16.1.0 compiles a C++23 source with the bound private glibc, gcc-runtime, and managed binutils
- [ ] `mcpp test` (full self-host build was terminated by an OOM kill while compiling `cli.m.cppm`)
- [ ] CI

### DIFF
--- a/src/build/ninja_backend.cppm
+++ b/src/build/ninja_backend.cppm
@@ -589,10 +589,17 @@ std::string emit_ninja_string(const BuildPlan& plan) {
     // The macOS initializer-ordering shim (#336) is a C translation unit, so
     // it needs the C driver bindings even in a project with no .c sources.
     const bool need_ios_init_shim = flags.needsStreamInitShim;
-    append(std::format("cxx       = {}\n", escape_ninja_path(flags.cxxBinary)));
+    auto compiler_command = [&](const std::filesystem::path& binary) {
+        const auto& dirs = plan.toolchain.compilerInvocationRuntimeDirs.empty()
+            ? plan.toolchain.compilerRuntimeDirs
+            : plan.toolchain.compilerInvocationRuntimeDirs;
+        return mcpp::platform::linux_::build_clean_ld_library_path_prefix(dirs)
+             + escape_ninja_path(binary);
+    };
+    append(std::format("cxx       = {}\n", compiler_command(flags.cxxBinary)));
     append(std::format("cxxflags  = {}\n", flags.cxx));
     if (need_c_rule || need_asm_rule || need_ios_init_shim) {  // asm_object drives the C compiler too
-        append(std::format("cc        = {}\n", escape_ninja_path(flags.ccBinary)));
+        append(std::format("cc        = {}\n", compiler_command(flags.ccBinary)));
     }
     if (need_c_rule || need_ios_init_shim) {
         append(std::format("cflags    = {}\n", flags.cc));
--- a/src/build/ninja_backend.cppm
+++ b/src/build/ninja_backend.cppm
@@ -597,6 +597,9 @@ std::string emit_ninja_string(const BuildPlan& plan) {
              + escape_ninja_path(binary);
     };
     append(std::format("cxx       = {}\n", compiler_command(flags.cxxBinary)));
+    // clang-scan-deps receives the driver after `--` as argv, not as a shell
+    // command. Keep a raw path for that interface; `$cxx` may start with env.
+    append(std::format("cxx_driver = {}\n", escape_ninja_path(flags.cxxBinary)));
     append(std::format("cxxflags  = {}\n", flags.cxx));
     if (need_c_rule || need_asm_rule || need_ios_init_shim) {  // asm_object drives the C compiler too
         append(std::format("cc        = {}\n", compiler_command(flags.ccBinary)));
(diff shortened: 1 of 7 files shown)


Sunrisepeak:
@ZheFeng7110 这个 `/etc/ld.so.preload` 的问题 是 libc 侧的问题。可以不用从mcpp侧强行 workaround, 并且从 mcpp 侧无法根本解决这个问题。目前已经从glibc侧解决了这个问题。可以使用最新版本进行尝试

```
xlings update mcpp -y

# 删除全局缓存目录
rm -rf ~/.mcpp

# 配置镜像(可选)
mcpp self config --mirror CN
```

然后验证。

ZheFeng7110:
好的，问题已解决。谢谢

PR #505, opened 2026-08-25
mirror: retry the completeness gate's GETs, and report the status it saw

Sunrisepeak:
The v2026.8.25.1 release failed twice on a single transient 502 from GitCode for an asset it had already mirrored — verified by hand as 5,772,395 bytes with the published sha256.

- both verification GETs retry with `--retry-all-errors` (plain `--retry` does not cover the transport-layer failures this path sees)
- `|| echo ERR` appended rather than replaced, so a 502 was logged as `502ERR` and the log could not be grepped for a status

Pure CI-tooling change; no effect on the build.

### DIFF
--- a/.github/tools/mirror_res.sh
+++ b/.github/tools/mirror_res.sh
@@ -133,7 +133,9 @@ done
 # The final completeness gate below still does FULL GETs.
 probe() { # host_path asset → 0 iff the object serves bytes
   local code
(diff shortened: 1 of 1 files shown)


Sunrisepeak:
Folded into #506 — the goal is one PR carrying the whole family.

PR #548, opened 2026-09-02
xlings: [xlings.workspace] is the one table, deps superseded, envs removed, and the install-time edge published (2026.9.3.1)

Sunrisepeak:
`[xlings]` is mcpp's surface for **xlings' local project mechanism** — the project `.xlings.json` that gives a directory its own environment. This brings the manifest into line with what that mechanism has.

Design record, with every measurement: `.agents/docs/2026-09-03-xlings-workspace-as-the-one-table.md`.

## What changes

**`[xlings.workspace]` becomes the one table.** An entry names a package and the version this project uses it at, and produces two projections: an install address (`[<ns>:]<target>[@<version>]`) for provisioning, and a resolution pin (`[<ns>:]<version>`) for the materialised file's `workspace` object — the file's two fields, which xlings reads in two different places.

The namespace may be written on either half; `"xim:picolibc-riscv" = "1.8.12"` and `picolibc-riscv = "xim:1.8.12"` are one entry. Writing it on the key needs quotes (a TOML bare key cannot hold a colon, measured: the unquoted form fails at the colon with a parser message that mentions nothing about namespaces). Writing it on both halves with different values is refused, as is naming one package twice. `""` means present with the version unconstrained — the spelling an authored project file already uses. Platform keys are xlings' own `linux` / `macosx` / `windows` / `default`, with `macos` kept as an alias.

**Why the split could not be defended.** `docs/05` §2.13 calls `[toolchain]` the shorthand and `[xlings.workspace]` the general form; the shorthand installs and the general form installed nothing. And nothing compared the two when both named one package.

**`deps` is superseded, still honoured, reported once** with the line to write instead. It is not refused: a refusal would reach a *dependency's* manifest, which a consumer that pinned an exact version of that package cannot edit. One package in both tables with two versions is a hard error — they are provisioned in order while the last wins the pin, so accepting it installs one and resolves the other.

**`envs` is removed.** It was materialised into `.xlings.json` and read by nothing: the two `envs` structures xlings has belong to a program's own shim record and to a SubOS's provider sections, and mcpp's own run environment comes from the runtime binding. The documentation described an effect that did not occur, which is why this is an error rather than a warning. No manifest in the ecosystem uses it.

**`mcpp emit xpkg` writes `xpm.<platform>.deps`.** Nothing emitted it before, so a package declaring an environment had its install-time edge written into the descriptor by hand — which is how `riscv-virt-rt` 0.3.0 shipped without the C library its own target row named. The declaration is kept unresolved per platform (`workspaceByPlatform`) because a descriptor has a block per platform while the host resolution has already discarded two.

**Documented: the three-row inheritance rule**, which has always held and which mcpp stated nowhere.

| The project declares | The version of a tool it did not name comes from |
|---|---|
| `[xlings.workspace]`, no `subos` | the machine's environment, with the project's entries over it |
| `[xlings.workspace]` and `subos = "<name>"` | that environment's own workspace; the machine's does not apply |
| neither | the machine's environment |

An `xlings use` inside the project outranks both, being the layer merged last.

## Withdrawn during implementation

The design's second change moved provisioning to project scope. The call site records that this was tried and measured: project scope puts payloads in `<proj>/.mcpp/.xlings/subos/_/usr/include` while `--sysroot` names the registry SubOS, so a declared dependency's headers were not found. The scope stays global; §3 of the design carries the withdrawal and what it leaves unsolved.

## Verification

e2e `331_xlings_workspace_one_table.sh` needs **no network on any shard**: every entry reaches the provisioning pass, and with `MCPP_NO_AUTO_INSTALL=1` that pass refuses and prints the addresses it was going to ask for, so the refusal is the assertion. It covers address assembly, both namespace positions, the unquoted-key error, `""`, the `deps` advisory and its suggested line, the two-table conflict, the `envs` refusal, and that a project declaring none of it hears nothing. It fails at its first assertion on 2026.9.2.1.

Unit: `test_manifest` +8. Full unit suite 97 passed. e2e 06, 93, 130, 131, 132, 134 re-run on this branch.

Version 2026.9.3.1; the bootstrap pin stays at 2026.9.2.1 until the release is indexed.


Sunrisepeak:
## Self-review, and what implementation changed

All 36 checks green. Five findings, four of them in this branch's own work.

**A nested `std::map` in an exported module truncates the BMI.** `XlingsConfig` first carried `map<string, map<string, string>>`. The module compiled and wrote an unreadable BMI: consumers failed with `failed to read compiled module cluster N: Bad file data` and `failed to load pendings for 'std::map'`, pointing at an unrelated file's ordinary `std::map` alias. The emitter needs the addresses and never the inner keys, so the field is a `vector` and says why. xlings' own source carries a note about the same GCC 16 shape with a different workaround.

**The design's second change was withdrawn against a measurement already in the code.** It moved provisioning to project scope; the call site records that this was tried and that project scope puts payloads in `<proj>/.mcpp/.xlings/subos/_/usr/include` while `--sysroot` names the registry SubOS, so a declared dependency's headers were not found. The scope stays global. §3 of the design carries the withdrawal and what it leaves unsolved.

**The removed key's first casualties were this repository's own fixtures.** e2e 88 and 205 declared `[xlings.envs]` and went red on CI rather than locally. That is the expected shape of removing a key that did nothing: the things that used it were the things that did not depend on it working. 88 now asserts the opposite — that `envs` is *not* materialized — and gains two criteria the merge introduced.

**One assertion could only fail on macOS, and did.** `Manifest.XlingsWorkspaceAcceptsPerPlatformValues` compared `host_platform_key()` against `"macos"`, and aligning that function with xlings' `macosx` made the comparison false on exactly one of the three hosts. The test was written in terms of the function so it would run everywhere; the literal on the other side is what defeated that.

**`macos`/`macosx` was a third copy of a two-line table.** The parser hand-rolled the fold and the host key beside a rule `modules/platform` already had. `mcpp::platform::xpkg_platform_key_for` is now the only place that knows; `TargetPlatform::for_os` is written in terms of it, the manifest reads it in both directions, and the host key comes from `xpkg_platform`. A unit test asserts the two spellings produce an identical declaration by resolving all three platforms explicitly rather than against the one it runs on.

## Deliberately not done

The three ecosystem packages are **not** migrated in this cycle. An older mcpp parses `[xlings.workspace]` and provisions nothing from it, so a package that migrated before its consumers moved would stop installing its emulator on every older engine, silently — worse than the advisory it would silence. §2.4 records the condition under which the migration is safe.

## Verification

e2e `331` needs no network on any shard: every entry reaches the provisioning pass, and with auto-install off that pass refuses and prints the addresses it assembled, so the refusal is the assertion. It fails at its first assertion on 2026.9.2.1.

Unit `test_manifest` +9; full unit suite 97 passed. Locally re-run: every e2e that touches `[xlings]` — 39, 52, 58, 60, 88, 131, 132, 133, 151, 200, 205, 220, 221, 237, 290, 327, 331 — plus 06 and 93 for the descriptor emitter.

PR #568, opened 2026-09-05
feat(clean): --stale removes the fingerprint directories no recorded build still uses (#565)

### LINKED ISSUE #565: `mcpp clean` has no middle ground between "keep everything" and "remove target/": fingerprint directories that are no longer current are never collected
Version: mcpp 2026.9.4.3 (release binary, linux-x86_64). Line numbers are against `main` at 3a9656e.

What a user observes. Every build writes into `target/<triple>/<fingerprint>/`, and when any fingerprint input changes the next build goes to a fresh directory and says so: `warning: fingerprint changed (a → b), full rebuild` (`src/build/execute.cppm:801-819`). That isolation is the right design; an old configuration can never contaminate an incremental build. But the directory the fingerprint moved away from is never touched again by anything. On a seven-member workspace after four days of iteration, one member had 17 fingerprint directories under `target/` across three triples, each around 4 GB, and the workspace as a whole carried about 219 GB of build output. Keeping only the newest directory per triple removed 70 directories and roughly 170 GB.

What exists today. `mcpp clean` removes `target/` wholesale, optionally the global build cache with it (`src/build/execute.cppm:2549-2568`). The global build cache, by contrast, has `mcpp cache prune` and `mcpp cache gc` with age and size budgets. `target/` has nothing between "do nothing" and "delete everything", and I could not find any code path that enumerates sibling fingerprint directories at all. Because the only remedy costs a full rebuild across every triple, people do not run it, and the stale trees accumulate silently.

What would answer it. A `mcpp clean --stale` (or a `gc` subcommand for `target/`, mirroring `cache gc`) that deletes every `target/<triple>/<fingerprint>/` not matching what the current configuration resolves to, with a `--dry-run` that lists candidates and their sizes. The build cache record already stores the current fingerprint per (target, profile, cache mode) entry (`BuildCacheEntry` in `src/build/execute.cppm`, from line 92; the comparison at `:809-815` reads exactly those three fields), so "current" is known without recomputing anything. Failing that, a one-line notice when a new fingerprint directory is created while older siblings exist, with their total size, would at least make the growth visible.

Workaround used today: keep the newest directory by mtime under each `target/<triple>/` and delete the rest. mtime is only a proxy for "current"; if the configuration changed after the last build, the newest directory is itself stale and the next build simply creates another one, which is what happened here on the first build after the cleanup.

yspbwx2010:
Closes #565.

`mcpp clean` had two settings: keep everything, or remove `target/` and rebuild every triple. This adds the middle one.

`mcpp clean --stale` removes every `target/<triple>/<fingerprint>/` directory that no recorded build considers current, and reports each one with its size. "Current" is what `target/.build_cache` already records: one entry per (target, profile) built recently, which is the same set the fast paths and `mcpp run` resolve to, so nothing that a plain `mcpp build` or `mcpp build --release` would reuse is touched. `--dry-run` lists the candidates and deletes nothing (it implies `--stale`). With no build record there is nothing to compare against, so the command refuses instead of guessing. `--stale` and `--bmi-cache` are rejected together, since the global build cache is shared across projects and already has `mcpp cache gc`.

The record is not the whole truth, and the code says so. It keys on (target, profile) while the fingerprint also folds in features; a `--no-cache` build writes no entry; and `mcpp test` builds through `run_tests`, which never writes one, so on a project that runs `mcpp build --release` and `mcpp test` the dev directory the tests live in is unrecorded. Recomputing fingerprints inside a clean command is not an option: `prepare_build` resolves dependencies and can reach the network. So the command borrows the rule `cache prune` already uses: an unrecorded directory written within `--older-than` (default `1d`, same duration syntax as `cache prune`) is kept and reported as kept; `--older-than 0` keeps none. Recording test builds in `target/.build_cache` would make the selection exact and is a separate change; I did not want to touch what the fast paths key on in this PR.

Two more limits are deliberate. Entries are matched by (triple directory, fingerprint) rather than by the absolute path the record stores, so a moved checkout is not read as "everything is stale". Only triple directories named by the record are visited, so `target/dist/` from `mcpp pack`, and any triple whose entry has been evicted from the eight-entry record, are left alone. Like `clean`, the command works on the package whose manifest is found from the current directory, so in a workspace it is run per member.

The `fingerprint changed` warning now ends with a pointer to the command, so the growth stops being invisible at the moment it happens.

Test: `tests/e2e/609_clean_stale.sh` checks the refusal before any build; builds dev and release (two current fingerprints under one triple); plants an old unrecorded directory and a freshly written one, and checks that `--dry-run` names the old one, reports the fresh one as kept, and removes nothing; that `--stale` removes exactly the old one; that `--older-than 0` then removes the fresh one and a bad duration is refused; that a second pass reports nothing stale; that `--dry-run` alone lists and that `--stale --bmi-cache` is refused; and finally rebuilds both profiles and asserts by mtime that no surviving binary was relinked. Docs: one line in both getting-started guides; changelog entry under Unreleased.

For scale: the workspace in the issue had 70 such directories, about 170 GB, identified with the mtime workaround described there; this command makes the same selection from the record instead of guessing from timestamps, and only falls back to age for directories the record cannot vouch for.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@
 
 ## [Unreleased]
 
+### `mcpp clean --stale`: 只清 target/ 里已无构建使用的指纹目录 (#565)
+
+每次配置指纹变化都会在 `target/<三元组>/` 下新开一个目录, 旧目录从不回收; `mcpp clean` 只有整删一档,
+代价是全量重编, 于是没人跑. `mcpp clean --stale` 以 `target/.build_cache` 里记录的 (三元组, 指纹) 为
+当前, 删掉记录过的三元组目录下其余兄弟目录并报告各自体积; 未被记录但在 `--older-than` (默认 1d) 之内
+写过的目录保留 (`mcpp test` 的构建不写记录). `--dry-run` 只列出. 没有构建记录时拒绝执行而不是猜; 记录之外的目录
+(如 `mcpp pack` 的 `dist/`) 不碰. `fingerprint changed` 的警告末尾现在附带这条
+命令, 让增长可见.
+
 ## [2026.9.5.3] — 2026-09-05
 
 ### 官方构建插件集中为一个包:`mcpp:plugins`
--- a/docs/00-getting-started.md
+++ b/docs/00-getting-started.md
@@ -131,6 +131,8 @@ human-readable.
 ```bash
 mcpp build              # incremental build
 mcpp clean              # clean target/
+mcpp clean --stale      # drop only target/<triple>/<fingerprint>/ dirs no build still uses
+                        # (--dry-run lists and deletes nothing; --older-than 3d keeps newer unrecorded ones)
 mcpp test               # compile and run tests/**/*.cpp — one binary per file,
                         # framework-agnostic (bare main, or gtest via [dev-dependencies])
 mcpp test <pattern>     # only tests whose name contains <pattern>
--- a/docs/zh/00-getting-started.md
+++ b/docs/zh/00-getting-started.md
@@ -122,6 +122,7 @@ workspace 中运行。插件稳定依赖进程退出码和生成的 `compile_com
 ```bash
 mcpp build              # 增量构建
 mcpp clean              # 清理 target/
+mcpp clean --stale      # 只删 target/<三元组>/<指纹>/ 下已无构建使用的目录 (--dry-run 只列出; --older-than 3d 保留更新的未记录目录)
 mcpp test               # 编译并运行 tests/**/*.cpp —— 每文件一个独立二进制,
                         # 框架无关(裸 main,或经 [dev-dependencies] 使用 gtest)
 mcpp test <pattern>     # 只运行名字包含 <pattern> 的测试
--- a/src/build/execute.cppm
+++ b/src/build/execute.cppm
@@ -25,6 +25,7 @@ import mcpp.build.backend;
 import mcpp.build.ninja;
 import mcpp.build.runtime_validation;
 import mcpp.bmi_cache;
+import mcpp.bmi_cache.maintenance;  // dir_size + human_bytes, for `clean --stale`
 import mcpp.manifest;
 import mcpp.source_kind;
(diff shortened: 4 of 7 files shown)


Sunrisepeak:
Reviewed and carried forward as #573, with your commit `79e26ca` and authorship preserved (plus a `Co-authored-by` trailer so the squash keeps it).

This branch lives in an organisation-owned fork, where GitHub does not honour "allow edits by maintainers" — pushing the review fixes here returns `Permission to cloud-teahouse/mcpp.git denied`, so they went to a base-repo branch instead.

The design holds up and is merged as designed: `target/.build_cache` as the single authority for "current", refusal instead of a guess when there is no record, and every limit erring toward a directory a rebuild can recreate. Three things changed:

- `mcpp clean --older-than 3d` set neither `--stale` nor `--dry-run`, so it reached `clean_project` and wiped all of `target/` — silently, and exactly what `--stale` exists to avoid. Any option of the mode now selects it.
- `--older-than -1s` (a typo for `1s`) parsed to a negative window through `parse_duration`, which keeps nothing; on the unpatched binary it exits 0 and deletes a directory written seconds earlier. Negative values are refused now.
- 609's closing assertion — the one that catches `clean --stale` destroying live build state — used GNU `stat -c` in a test that declares no capabilities, so it ran in the macOS suite where BSD stat rejects `-c`. `stat ... | sort` inside `$(...)` reports `sort`'s status and there is no `pipefail`, so both sides compared equal as empty strings and it printed `OK`. It now uses the portable idiom from 142/183.

Thanks for the feature — the reasoning in the `clean_stale` comment is what made the review quick.

Sunrisepeak:
Merged as 090c0163 — your commit 79e26ca is in it, and the squash carries `Co-authored-by: Cloud_Yun <yunfeng66645@gmail.com>`. Closing here because this branch is in a fork the repository cannot push to, so the review fixes could not land on it. Thanks for the feature.

PR #581, opened 2026-09-06
fix: a dependency's [feature-xlings] tools must be installed before its build.mcpp runs

Sunrisepeak:
The root's `[xlings.workspace]` declarations are provisioned early. The GRAPH's
were provisioned about 1700 lines further down, after every build.mcpp had
already run. So a package that declares a tool under the feature that needs it
and then asks for it with `xpkg_dir` worked as the ROOT and was refused as a
DEPENDENCY -- with the very declaration it had already made quoted back at it.

Measured on the published `ggml-org:llamacpp@b10069.2`, which is exactly this
shape:

    [feature-xlings.backend-vulkan]
    "xim:shaderc" = "2026.3"

A consumer that writes the one line its README asks for could not build on a
clean machine:

    error: dependency 'llamacpp': build.mcpp exited with 2 (build aborted):
    ggml-org:llamacpp backend-vulkan needs glslc. Declare it: ...

With an isolated MCPP_HOME the unfixed engine pulled twenty-four xim payloads
for that graph and not shaderc; the fixed one pulls it and the consumer builds,
1921 shaders compiled.

WHY NOTHING CAUGHT IT. Once the tool is in the registry for any reason, and
building the package itself puts it there, `xpkg_dir` finds it and the ordering
stops mattering. llama.cpp-m's own CI builds it as the root; every machine that
had touched the package had the payload. Only an empty registry can see this,
which is what the sandbox run is for.

The split is now computed once, beside `activeFeaturesByPackage`, and read by
both the early provisioning pass and the later record-writing pass, so the two
cannot disagree about what "the graph declared" means. The late provisioning
call is kept: the stamp is keyed by content, so it is normally a hit, and it
remains the site that makes the record true if the early pass did not run.

e2e 618 is the criterion, and its first version was a false green worth
recording: it used `xim:ninja`, which mcpp installs to run its own builds, so
`xpkg_dir` answered on the broken engine too. It now uses the tool the defect
was actually found with.


Sunrisepeak:
Two-engine verification, isolated `MCPP_HOME` on both:

```
NEW engine (this PR):     PASS — tool at .../xim-x-shaderc/2026.3
OLD engine 2026.9.6.2:    FAIL — "not provisioned before its build.mcpp ran"
```

The first version of e2e 618 used `xim:ninja` and passed on BOTH engines: mcpp installs ninja to run its own builds, so `xpkg_dir` answered regardless of whether the feature had been provisioned. A criterion that selects an object present for an unrelated reason cannot fail. It now uses the tool the defect was found with.

Issue #599, opened 2026-09-09
233_bench_matrix: matrix.json's hub does not exist in the pinned tree, and the check only runs where submodules are initialised


Sunrisepeak:
Found while running the full e2e suite locally (332 passed, 1 failed, 35 skipped). Unrelated to the change that surfaced it.

## What fails

```
FAIL: bench/matrix.json
  linux/gcc/mcpp-2026.8.11.3:   hub='modules/platform/src/platform.cppm' does not exist in the pinned tree
  linux/clang/mcpp-2026.8.11.3: hub='modules/platform/src/platform.cppm' does not exist in the pinned tree
  windows/clang/mcpp-2026.8.11.3: hub='modules/platform/src/platform.cppm' does not exist in the pinned tree
```

`bench/projects/mcpp/mcpp-2026.8.11.3` is checked out at `a749e9f` (= `v2026.8.11.3`), and `modules/platform/src/platform.cppm` does not exist in that tree — the file lives at `src/platform.cppm` there. The path in `matrix.json` was written for the CURRENT layout after the module move.

The test's own message says what that costs: *"every scenario that perturbs it would be reported `skipped` and the job would still pass"*.

## Why CI is green

The check skips a workload whose tree has no `mcpp.toml`, which it reads as "submodule not initialised":

```python
if not os.path.isfile(os.path.join(tree, "mcpp.toml")):
    ...
    uninit.add(c["project"])
    break
```

Only the bench workflow checks submodules out, so on every other job the three cells are skipped and the test passes. On a developer machine with submodules initialised, it fails — which is where it was found.

## What a fix has to decide

Not just which path to write. The pinned tree is a *historical* mcpp, so a hub path that is correct today is wrong for it by construction, and the next module move breaks it again. The options are a per-project hub (the pinned tree's own path), or a check that resolves the hub against each tree rather than against one string.

Whichever is chosen, the `uninit` branch is worth revisiting in the same pass: a check that is silent on every CI job except one is a check whose failures are found by hand.

Sunrisepeak:
Fixed in #607, released in 2026.9.11.2. The hub path in `bench/matrix.json` now names the pinned tree's own layout (`src/platform/platform.cppm`). The uninitialised-submodule branch fails under `CI=true` instead of printing a note. Confirmed on main at c688fcab.

PR #608, opened 2026-09-11
docs(record): ecosystem sign-off for the 2026.9.11.2 batch

Sunrisepeak:
Section 9 of the six-issue record: the ecosystem-level review the goal asks for, written after the work landed and against what shipped rather than against the plan.

Four findings that belong to the ecosystem rather than to any one change:

- a **proxy check refused correct output twice in two repositories** on the same day (the MSI size ratio, and earlier the AppImage size bound) -- both replaced by opening the artifact;
- the **#599 fix reproduced the shape of the defect it fixed**, and the macOS lane caught it;
- the **ecosystem rule cost one new payload** and otherwise only declarations, with `xim:python`'s single-arch table turning out to be the stated reason emsdk could not declare an interpreter;
- the rule **composes with the licence rather than above it**, and getting that order wrong is asymmetric -- GitCode assets cannot be deleted.

Also states what is still open with its size named: the four target rows are the next PR, and `xim:wix` is a gap with a known shape.

Documentation only.

### DIFF
--- a/.agents/docs/2026-09-11-distribution-plugins-and-platform-decomposition.md
+++ b/.agents/docs/2026-09-11-distribution-plugins-and-platform-decomposition.md
@@ -1116,6 +1116,22 @@ unpersisted declaration would be absent exactly when a user names a format.
 carries no such line and the program that wrote it could not emit one, so
 replaying it yields what that program said.
 
+Measured across two real binaries rather than only in a unit test, because an
+absent-tolerance claim is about what a *previous version* wrote:
+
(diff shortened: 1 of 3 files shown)


Sunrisepeak:
Folded into the single mcpp PR for this goal, per the goal's one-PR-per-repo rule. The ecosystem sign-off (section 9) and the two record corrections are now commits on the branch that also carries the target-row work, released as one version (2026.9.11.3) rather than as a separate documentation PR.

PR #670, opened 2026-09-18
release.yml: an unmatched CHANGELOG version fails the release, not a placeholder

Sunrisepeak:
## Summary

A defect the c-environment wave (#668) revealed in the release machinery, not
part of that wave's own implementation — filed separately on purpose.

`v2026.9.18.1`'s GitHub Release published with its notes body reading
literally `(no CHANGELOG entry found for 2026.9.18.1)`. The "Extract release
notes from CHANGELOG" step in `release.yml` scans `CHANGELOG.md` for a
`## [<version>]` heading; #668's entries had been written under
`## [Unreleased]` only, so the scan found nothing, wrote that placeholder
string, and let the release publish anyway. A release is outward-facing and
permanent — it should not be possible to publish one whose notes silently
came out empty.

## Change

Replaces the placeholder-and-continue with a hard failure:

```bash
if [ ! -s dist/RELEASE_NOTES.md ]; then
  echo "::error::CHANGELOG.md has no '## [$VERSION]' section — refusing to publish a release with empty notes. Add that section (docs/92-release.md) and re-run, ..."
  exit 1
fi
```

The job stops before "Create GitHub Release" runs, so nothing publishes with
empty notes. No redesign of the extraction logic itself — same `awk`, same
heading format, same file.

## Verification

**This repo has no workflow-execution harness** (no `act`, no dry-run job for
`release.yml`) to exercise this step directly in CI, so I verified it the only
way available: extracted the exact shell block and ran it standalone against
two synthetic `CHANGELOG.md` fixtures.

- A version WITH a matching `## [<version>]` section → notes are printed,
  exit 0.
- A version with NO matching section → the `::error::` line is printed,
  exit 1.

```
=== Case A: matching section present ===
--- RELEASE_NOTES.md ---

### Something happened

- a bullet
exit code: 0

=== Case B: no matching section ===
::error::CHANGELOG.md has no '## [8.8.8]' section — refusing to publish a release with empty notes.
exit code: 1
```

Also ran the repo's existing static check against the changed file:

```
$ tools/lint-ci-assertions.sh .github/workflows/release.yml
ci-assertions: no findings in 1 workflow file(s)
```

`CHANGELOG.md` already carries a `## [2026.9.18.1]` section (fixed separately,
after #668 merged), so this change does not retroactively fail the release
that already shipped — only a future one whose CHANGELOG section is still
missing at tag time.

## Not in scope

Per the coordinator's explicit instruction: this PR does not redesign the
notes-extraction logic, and does not invent a workflow-execution test harness
that does not already exist in this repo — the manual verification above is
reported as manual, not dressed up as an automated test.


Sunrisepeak:
Merged by admin override because the ruleset requires two approvals and only one operator is present.

PR #672, opened 2026-09-18
c-abi probe: strip Windows host macros; wchar always emitted for freestanding (2026.9.18.3)

Sunrisepeak:
## What this PR fixes

The c-abi probe fails on two real pairings today, both on Windows
hosts cross-compiling to a non-Windows target where the host's
toolchain leaks macros the target's --target= substitution does not
strip:

  openkal-llvm-runtime#24:
    Windows host × riscv64-none-elf (freestanding)

The probe reports:

  __SIZEOF_WCHAR_T__  declared 32  measured 16
  _WIN32              declared undefined  measured defined

Both are defects at the engine boundary, not in either package's
manifest, and neither is reachable on Linux or macOS hosts (both pass
today). The structural root causes are:

1. clang on a Windows host defines _WIN32 in its preprocessor output
   for ANY --target= triple (hosted --target= rewrites host macros;
   freestanding --target= does not). The probe has no way to
   compensate.

2. musl's [c-abi] declaration says wchar = 32, because musl's
   wchar_t is 32 bits on every target it covers. For freestanding,
   the toolchain's host-contaminated default controls the actual
   compile (16 on Windows hosts), and the engine was relying on a
   "freestanding skips the wchar flag" rule that does not hold on
   Windows hosts.

## What this PR does

c-abi probe gains a `hostStripMacros` parameter (defaulting to {} for
backwards compat) that the build layer can use to insert -U<name>
tokens before -E -dM. The cache key folds the strip list in so two
probes that differ only in strip list do not share a slot.

The build layer (prepare.cppm) passes `_WIN32`, `_WIN64`,
`__MINGW32__`, `__MINGW64__` to hostStripMacros on every c-abi probe
it runs, regardless of target. This is the four names the build's
own measurements have surfaced as Windows-host leaks; the parameter
is a list (not a fixed set) so a future host whose compiler leaks a
different macro can be handled the same way without further changes
to this module.

The wchar realisation (cenv.cppm) is restructured so a freestanding
target gets -fno-short-wchar unconditionally for wchar=32 (and
-fshort-wchar for wchar=16); the hosted branch keeps the old
"compare against native_wchar_bits(os)" gate. Without this, the
Windows-host × freestanding pairing still leaks 16-bit wchar_t into
the compile the engine then measures.

## Versioning

This is 2026.9.18.3. mcpp.toml and modules/versioning are bumped;
docs/22 and CHANGELOG are updated.

## Measured

openkal-musl#37 (this fix's 2026.9.18.3 binary, locally built):
5/5 CI green, including cross-link for the other system (the
freestanding QEMU path).

openkal-llvm-runtime#24 (this fix's 2026.9.18.3 binary):
- build the runtime: PASS
- the artefact built on Linux runs on macOS: PASS
- the artefact built on Linux runs on Windows: PASS
- macOS host reaches every target: PASS (was already passing on .2)
- Windows host reaches every target: PASS (was FAIL on .2)

mcpp's own unit tests for cenv and cenv_probe (test_cenv.cpp,
test_cenv_probe.cpp, added by this PR): PASS locally on x86_64-linux.
The new test_cenv_probe.cpp covers both the strip-list folding into
the cache key and the four Windows-host macros' effect on a probe
that asks for them undefined.


Sunrisepeak:
Superseded by #673 — same fix, plus an additional test refinement commit. All openkal ecosystem CI was triggered against this branch's content (the rebased 3-commit series e2ed1e11 / f661260c / 534b1ee2 carries the same fix as #673's first three); #673 adds a fourth test refinement. Closing this in favor of #673 for clarity.

PR #681, opened 2026-09-20
2026.9.21.2: the macros mcpp owns are spelt in upper case, and the borrowed name is withdrawn

Sunrisepeak:
`__mcpp_target_<os>__` becomes `__MCPP_TARGET_<OS>__` and `__openkal__`
becomes `__OPENKAL__`. The spelling is still the triple's own `os` field, now
upper-cased; the engine still learns no operating-system name.

THE CONVENTION SPLITS BY WHAT A NAME IS, NOT BY WHO WRITES IT. A vendor or
product name is upper --- `__APPLE__`, `_WIN32`, `__MINGW32__`, `__GNUC__`. A
kind-of-system name is lower --- `__linux__`, `__unix__`, `__gnu_linux__`.
Every row this engine OWNS is of the first kind: it names mcpp, or it names
openkal. The kind-of-system question is answered by `__linux__` and its
family, which mcpp SUPPLIES rather than owns and which therefore keep their
lower-case spelling, for that exact reason.

The lower-case spelling shipped in 2026.9.21.1 on the reverse reading: that
these names sit beside `__linux__` in real guards, so matching it was
consistency. THAT CONFUSES ADJACENCY WITH KIND. `__APPLE__` sits in those
same guards and is upper, because it belongs to somebody.

`__MCPP_` is a prefix, not the whole rule: `__OPENKAL__` is owned and names
*openkal* rather than mcpp. The assertion in `test_predefines.cpp` therefore
checks the SPELLING CONVENTION --- upper case, `__`-wrapped --- instead of
the prefix, which would have judged `__OPENKAL__` a violation when it is not.

WITHDRAWING A MACRO IS DECIDED BY A COUNT, NOT BY READING

An entry in this contract is a published interface, and withdrawing one is
SILENT: a `#if` selects the other branch and compiles. No mechanism available
to a build tool makes that loud. So a withdrawal is decided by a measurement
that ENUMERATES READERS, and what it finds sets how many steps the withdrawal
takes. The two this project has performed came out differently:

    withdrawn                              readers found            steps
    __CYGWIN__                             4 third-party members,   3
                                           + 6 sites in the two
                                           headers we INSTALL
    __mcpp_target_<os>__, __openkal__      none                     1

The second row's denominator is every repository of this ecosystem, swept by
file type: no source file and no manifest reads either name, and the only
occurrences are this engine's emitter, its tests, and prose. Both names were
invented here, so no upstream code can hold one; exposure was three days for
`__openkal__` and a single release for the target macro. The rule did not
change between the two rows. The count did.

A new assertion takes its denominator from the target registry rather than
from a list written beside the test: every canonical triple is parsed and its
emitted macro checked to be a valid identifier. The spelling comes from the
`os` field, so an `os` carrying a dot or a version suffix would produce a
macro no compiler accepts, and the failure would land in a user's build.

__CYGWIN__ IS WITHDRAWN --- THE LAST STEP OF A THREE-REPOSITORY SEQUENCE

The Windows `presents = "posix"` realisation now adds `-U__CYGWIN__` and
`-U__CYGWIN32__` to the compile line. `--target=x86_64-pc-cygwin` stays: it
is what supplies `__unix__` and suppresses `_WIN32`, which is what presenting
POSIX means. Only the borrowed NAME was unwanted.

  1. 2026.9.21.1 defined mcpp's own name beside the borrowed one; this
     release re-spells it, while it still has no consumer. Purely additive.
  2. `openkal-musl@0.19.0` and `openkal-llvm-runtime@0.14.0` read the new
     name and keep `|| defined(__CYGWIN__)`, so they build on an engine from
     either side of this change. PUBLISHED BEFORE THIS RELEASE.
  3. this release stops defining the borrowed name.

THE ORDERING IS MEASURED, NOT ASSERTED. It holds between repositories, so no
test in this one can check it. Building an openkal program for
`x86_64-windows-gnu` against the PUBLISHED `openkal-llvm-runtime@0.13.0` on
this engine fails on libunwind's two `static_assert`s --- `x86_64 registers
do not fit into unw_context_t`, `UnwindCursor<> does not fit in
unw_cursor_t`. That is step three taken first, and it is why step two ships
first. `setjmp.h` is the silent half: its own comment says a mismatch is
reported by nothing until the record overruns.

THE RESIDUAL WINDOW IS NAMED RATHER THAN CLAIMED AWAY. A project pinning
`openkal-musl` at 0.18.0 or earlier EXACTLY, and upgrading past this release,
gets that silent `#else`. Moving the index's `latest` onto 0.19.0 first keeps
the window to exact pins; nothing available here closes it, because the
engine cannot know which macros a package's installed headers read.

Bootstrap pin advances from 2026.9.20.1 to 2026.9.21.1.


Sunrisepeak:
## Ecosystem review before merge

Per repository, what this release does to something already published.

### Published packages

| package | effect | measured |
|---|---|---|
| `openkal-musl` ≤ 0.18.0 | **breaks** on a Windows target presenting POSIX: `bits/setjmp.h` reads only `__CYGWIN__`, falls to `#else`, short `jmp_buf`, reported by nothing | yes — the sibling defect in libunwind fires two `static_assert`s on the same build |
| `openkal-musl` 0.19.0 | unaffected: reads `__MCPP_TARGET_WINDOWS__` with `\|\| defined(__CYGWIN__)` beside it | yes, cross-verified against this branch |
| `openkal-llvm-runtime` ≤ 0.13.0 | **breaks**, loudly: `x86_64 registers do not fit into unw_context_t` | yes, built on this engine against the published 0.13.0 |
| `openkal-llvm-runtime` 0.14.0 | unaffected | yes — five jobs green against this branch, including the Windows-host row of `host-dimension` |
| every other package in the index | untouched: no manifest key changed, no descriptor grammar changed | `check_version_pins`, and 231 descriptors already parse |

**The ordering is the mitigation and it is not optional.** 0.19.0 and 0.14.0 are
published and registered; the index's `latest` for both has moved. This release
goes out after that, not before.

### What is not mitigated, and is named rather than claimed away

A project pinning `openkal-musl` at 0.18.0 or earlier **exactly**, and upgrading
the engine past this release, gets the silent `#else`. No mechanism available
here closes it: the engine cannot know which macros a package's installed
headers read. Recorded in `docs/22` and in `cenv.cppm` at the emission site.

### The renamed macros

`__mcpp_target_<os>__` and `__openkal__` stop being defined. Enumerated across
every repository of this ecosystem, by file type: **no source file and no
manifest reads either name**; the only occurrences were this engine's emitter,
its tests, and prose. Both names were invented here, so no upstream code can
hold one, and exposure was three days for one and a single release for the
other. A withdrawal argued from anything other than that count is the argument
this project already got wrong once, in the first shape of this same branch.

### Index

No new manifest key, so `min_mcpp` does not move and no client loses the index.
`latest_mcpp` moves with the release as usual.

### Build cache

The macro rename changes field 7 of the fingerprint. It costs nothing: field 8
is the mcpp version, so every release already invalidates every entry.

### What this release is expected to FIX

Four members of the compatibility measurement — `archive`, `sqlite3`,
`mimalloc`, `c-ares` — stop at `#include <windows.h>` reached through
`#if defined(_WIN32) || defined(__CYGWIN__)`. The measurement that says so is
the one that moves `pins.toml` onto this release, and it is the criterion this
change is accountable to. It has not run yet; the figure will be reported
there, not here.

